PR #15, opened 2019-03-13
(WIP)Real time filtering of capacitive probe

### LINKED ISSUE #14: Real time filter of capacitive probe output
Complete description of the problem [here](https://docs.google.com/document/d/1TXaYTfHYrfTIfu60gRR-6B_dla-Ja4RMTo_o7mJaNeg/edit).

Let's use this thread to follow code implementation

### LINKED ISSUE #19: Reference to non-existent field 'probeType'
plugging capacitive sensor on my laptop, then running this:
~~~
>> P = ProbeTracking();
Reference to non-existent field 'probeType'.

Error in ProbeTracking.declareprobe (line 966)
    switch AuxSpecs.probeType

Error in ProbeTracking (line 126)
    [ Aux.Source, Aux.Specs ] = ProbeTracking.declareprobe( Specs ) ;
~~~

things should probably get configured under `/dev` but this needs documentation (hence this issue)

Aureliengilliot:
This PR implements a better way to filter the capacitive coupling probe. Fixes #14

TODO:
- Cleaning of useless files.
- Add README [here](https://github.com/neuropoly/realtime_shimming/tree/master/RespiratoryProbe): to explain how to use those files.

How to test:
- filter_signal('capacitive', 'test/capacitive_probe_baseline-capacitor.mat')
- Dummy sensor signal for testing: https://drive.google.com/drive/folders/1wRUHbCtusUH1SMNaKuAw6zsN9k6txEU8

Other improvements:
- Reference to non-existent field 'probeType'. Fixes #19


Aureliengilliot:
Tried implementing the fitting procedure on the signal with complete periods, but it fails because the signal is unbalanced:
untitled

Next thing to try is to average the signal within period, and to fit that.

jcohenadad:
OK, for now i've opted for an approach where the poly degree is updated after X timepoints. This is very empirical but should work OK for now. We can improve it in the future.

jcohenadad:
@rtopfer @alexfoias please let us know how this function should be called by the rest of the code, so we can finalized i/o stuff

rtopfer:
It should be called (and possibly embedded as a subfunction) within [ProbeTracking.getupdate()](https://github.com/neuropoly/realtime_shimming/blob/84821815ebdf9ba0c171660900b3263c51b0d6f4/RespiratoryProbe/ProbeTracking.m#L191)

Nicolas-Pinon:
Simple filtering did not give better result than the polynomial fitting (frequency tuning of the filter was a pain)
To overcome the problem of the low frequency part of the signal varying too much, the fitting of the polynomial curve is now only updated **every period** of the signal instead of every time point, this way there is less small variations.
This also make sense because there is no point in updating the fit every point since we are not interested in the high frequency variations for the fit.

jcohenadad:
@Nicolas-Pinon thank you for the investigations. Would it be possible to point to a dataset and procedure for testing (in order to review this PR). Thanks!

Nicolas-Pinon:
> @Nicolas-Pinon thank you for the investigations. Would it be possible to point to a dataset and procedure for testing (in order to review this PR). Thanks!

Right now I will be going to my brass band rehearsal, could that be a work for @Aureliengilliot ?
I'll help him if there is trouble, but by looking at the before/after my fix, I'm pretty confident in it.

jcohenadad:
> looking at the before/after my fix, I'm pretty confident in it

documenting a testing framework will enable to derive unit testing in the future, and make sure the algorithm will still work until the years 2050

alexfoias:
@Nicolas-Pinon We have tested the implementation in real time. The current version (afdc8aafade3a629b239f769269cb9bfef77039f)(fitting based on each period) doesn't seem to work. 

You can test the algorithm with this dataset. 
[20190606_11-33-03AM_VariablesDump_au.mat.zip](https://github.com/neuropoly/realtime_shimming/files/3262421/20190606_11-33-03AM_VariablesDump_au.mat.zip)

UPDATE: I added the git version. 

UPDATE: I used get_capacitive_probe_signal from alfoi/CapacitiveProbe branch. In order to load the above dataset I changed line 8 to `load('../test/20190606_11-33-03AM_VariablesDump_au.mat');`

jcohenadad:
guys, before moving forward, [as I already asked](https://github.com/neuropoly/realtime_shimming/pull/15#issuecomment-498839493) please document how the testing is done (if each of you test in a different way, we are not making progress). So, on top of this PR, document:
- what command to run in order to run the function with a specified mat file
- list a couple of datasets used for validating this PR

when you test and report in this PR, always indicate the git version so we can compare across previous implementations

alexfoias:
I documented more the previous test on the initial comment (https://github.com/neuropoly/realtime_shimming/pull/15#issuecomment-499559093).

jcohenadad:
> on the initial comment

what is the initial comment?

alexfoias:
This comment: https://github.com/neuropoly/realtime_shimming/pull/15#issuecomment-499559093

jcohenadad:
@alexfoias can you add an optional input param to the function `get_capacitive_probe_signal()` so we don't have to modify the code itself for debugging, but instead do something like: `get_capacitive_probe_signal(file_recording=blablabla.mat)`

then, please, update the [top comment](https://github.com/neuropoly/realtime_shimming/pull/15#issue-260869037) to explain how to test the stuff and list a couple of mat files to download for testing.

alexfoias:
Implemented the feature here: https://github.com/neuropoly/realtime_shimming/commit/d37d330a1d0ced95e637bbbf42293706ed0796ea

EDIT: I updated the [top comment](https://github.com/neuropoly/realtime_shimming/pull/15#issue-260869037) to explain how to use the code.

jcohenadad:
@rtopfer you mentioned you worked on this files recently to implement the "watchdog feature" (i.e. acquiring from one station, and creating the recording from another station). What branch are these features implemented in?  I'd like to make sure this PR is based on the latest code

rtopfer:
@jcohenadad the daemon recording functionality using [ memmapfile()](https://www.mathworks.com/help/matlab/ref/memmapfile.html) is on the master branch. The polynomial fitting, filtering, etc. would ideally be implemented within [ProbeTracking.processupdate()](https://github.com/neuropoly/realtime_shimming/blob/86b198fe9ba24b79569b2f0d81d8f17b2c227f3d/RespiratoryProbe/ProbeTracking.m#L278)

jcohenadad:
thank you @rtopfer. This PR needs to be updated then. Also, I strongly recommend to create an external function which would only deal with the detrending, and which would be called [there](https://github.com/neuropoly/realtime_shimming/blob/86b198fe9ba24b79569b2f0d81d8f17b2c227f3d/RespiratoryProbe/ProbeTracking.m#L302) (for code clarity).

alexfoias:
@Aureliengilliot @jcohenadad I added the baseline measurement [here](https://github.com/neuropoly/realtime_shimming/tree/alfoi/CapacitiveProbe/test)

RespiratoryProbe/ProbeTracking.m
```diff
@@ -51,16 +51,24 @@
...
+    if exist('Aux')
+    % Nonurgent TODO:
+    %   This form of ProbeTracking() initialization/construction is not to
+    %   be called by a user, but rather, from the ProbeTracking() constructor
+    %   itself, making it better suited as a 'private' constructor. Apparently
+    %   Matlab does not permit this. There is a work-around described here:
+    %   https://stackoverflow.com/questions/29671482/private-constructor-in-matlab-oop
+        Aux.launchrecordingdaemon() ; % runs continuously in background
+        return;
+    else
```
jcohenadad, line 130: @rtopfer when launching `ProbeTracking(recording.mat)`, `Aux` does exists and hence it does not go inside this else. What did you have in mind here?

RespiratoryProbe/ProbeTracking.m
```diff
@@ -1,22 +1,22 @@
...
 %       .Log
-%
 %       .Source    
-%
 %       .Specs
 %
 % =========================================================================
-%
-% =========================================================================
-% Updated::20190528::ryan.topfer@polymtl.ca
```
jcohenadad: no date-- GH is there for that

RespiratoryProbe/ProbeTracking.m
```diff
@@ -51,16 +51,24 @@
...
+    %   This form of ProbeTracking() initialization/construction is not to
+    %   be called by a user, but rather, from the ProbeTracking() constructor
+    %   itself, making it better suited as a 'private' constructor. Apparently
+    %   Matlab does not permit this. There is a work-around described here:
+    %   https://stackoverflow.com/questions/29671482/private-constructor-in-matlab-oop
+        Aux.launchrecordingdaemon() ; % runs continuously in background
+        return;
+    else
+        Specs.state = 'inert' ;
+        error('TODO: enable debug mode: previous (raw) recording is loaded and can be processed as if it were a new recording...')
```
jcohenadad, line 132: @rtopfer i am not familiar with the deamon procedure (and there is no explanation in your code as to how to set this up)-- so any tips here would be appreciated (what function to call? etc.)

RespiratoryProbe/ProbeTracking.m
rtopfer: ProbeTracking() receives an input filepath (leading to .mat) in 2 foreseeable cases:
1. this debugging feature we're implementing now, in which a user directly calls ProbeTracking( _filepath_ ) ;
2. when ProbeTracking() is called by the daemon matlab session in the dual-session mode. That also involves a .mat file, but in the first case that .mat shouldn't contain a variable called 'Aux' (and therefore, exist('Aux') would return false). Clearly that would be violated if for some reason the user decided to name another variable 'Aux'...  I will add other check conditions.

jcohenadad: ok, while you do that, could you please also comment the file header (i've started doing it, but since my understanding of this code is limited, you'd better update my comment)

RespiratoryProbe/ProbeTracking.m
rtopfer: As it stands, ProbeTracking.declareprobe() looks for the device ; if it finds it, the daemon session is launched. (i.e. there is no function to call). 
For the debugging case, we don't need a daemon session (I was suggesting Specs.state = 'inert' as a sort of flag to avoid launching a 2nd session, but this flag could be coded differently (e.g. Specs.state = 'debug'))...

jcohenadad: ok, but the `error()` should ultimately be replaced with "something";  this "something" being what is typically launched by the deamon, so any tips about the workflow of the code would be welcome (could be in the header of the file)

p.s. `Specs.state = 'debug'` sounds better (unless there are other reasons than debugging for it to be inert)

rtopfer:
@jcohenadad disabling daemon mode:

_Specs.isRecordingDaemonEnabled = false ;
P = ProbeTracking( Specs ) ;_

Note that _P.recordphysiosignal( )_ in this mode will always begin an entirely new recording (i.e. fitting routines start from scratch)

jcohenadad:
@rtopfer when stopping and re-running `P.recordphysiosignal`, i get the following:
~~~
>> P.recordphysiosignal
Error using serial/fopen (line 72)
Open failed: OBJ has already been opened.

Error in ProbeTracking/beginrecording (line 189)
    fopen( Aux.Source ) ;

Error in ProbeTracking/recordphysiosignal (line 606)
isRecording = Aux1.beginrecording() ;
~~~
i need to close/reopen matlab. Any clue how to make it easier?

**UPDATE 2019-06-14 13.13**: Nevermind, I cannot reproduce that error anymore (dunno why)

**UPDATE 2019-06-14 13.15**: For the record: This bug happens when the session crashes (i.e. recording not properly terminated, I guess).

**UPDATE 2019-06-14 13.26:** To list opened items: `instrfindall`. To close: `fclose(instrfindall)`.

PR #18, opened 2019-06-12
Added support for shim board revision C control

alexfoias:
This PR aims to establish the teensy code for rev C of the shim amplifier board.

Code is based on Jason's code located on Gdrive/NeuroPoly/Projects/ACDC_SPINE_7T/Jason_ShimBoard/teensy/revCRun_Blocks_dummy.zip


Shim_revC/ShimOpt_revC.m
```diff
@@ -0,0 +1,257 @@
...
+Params.currents = repmat( [0.5 -0.5], [Params.nChannels 1] ) ; , % [units: A]
+
+% magnitude threshold to define phase unwrapping region
+Params.threshold       = 0.001 ;
+
+% 2 columns: [ MAG | PHASE ] ;
+Params.dataLoadDirectories = cell( Params.nEchoes, 2, Params.nCurrents, Params.nChannels ) ;
+
+tmp = { ...
+    '/Users/ryan/Projects/Shimming/Acdc/Calibration/data/acdc_60p/33-fm2d_CH1_+0.5A/' ;
```
jcohenadad, line 187: hardcoded path-- will it work?

rtopfer: Few options:
1. Hardcode, but use Duke paths (this seems most reasonable to me for now, until we have "the best" shim reference maps such that we don't care to reacquire them)
2. Include all the DICOMs in the Google drive (each data set could be ~1Gb, and we may want several data sets...)
3. Host them somewhere else...
That being said, why is there a ShimOpt "_revC"?

jcohenadad:
@rtopfer can we consider merging this branch?

alexfoias:
@jcohenadad I think we can merge this branch, we've tested it in the past. If we want to optimize more we can do it later.

jcohenadad (approved):

PR #26, opened 2019-09-03
(WIP)Few changes in shimdemo file

jcohenadad:
Changes include:
- Updated link to resp. probe mat file

### DIFF
--- a/example/shimdemo.m
+++ b/example/shimdemo.m
@@ -66,11 +66,14 @@ Field = FieldEval.modelfield( Fields ) ;
 
 
 %% -----
-% create a ShimOpt object corresponding to the hardware
+% Create a ShimOpt object corresponding to the hardware
 Params = [] ;
 Shims  = ShimOpt_Greg( Params, Field ) ;
 
-% define the target region for shimming with a binary mask (same size as the Matlab variable: Field.img)
+% The Shims.Aux contains the ShimOpt object of the Siemens Gradient+Shim coils (called: ShimOpt_IUGM_Prisma_fit), which is used for joint optimization.
+% Note: If you instanciate Shims directly using using ShimOpt_IUGM_Prisma_fit, then the .Aux field will be empty.
+% 
+% Define the target region for shimming with a binary mask (same size as the Matlab variable: Field.img)
 %
 % by default, it will be the intersection of Field.Hdr.MaskingImage (where
 % reliable field values exist) and Shims.getshimsupport() (where shim reference
--- a/example/shimdemo.m
+++ b/example/shimdemo.m
@@ -70,7 +70,7 @@ Field = FieldEval.modelfield( Fields ) ;
 Params = [] ;
 Shims  = ShimOpt_Greg( Params, Field ) ;
 
-% define the target region for shimming with a binary mask (same size as Field.img)
+% define the target region for shimming with a binary mask (same size as fieldB0.nii)
 %
 % by default, it will be the intersection of Field.Hdr.MaskingImage (where
 % reliable field values exist) and Shims.getshimsupport() (where shim reference
--- a/example/shimdemo.m
+++ b/example/shimdemo.m
@@ -18,7 +18,7 @@ Probe = ProbeTracking( '~/Matlab/shimming/example/acdc_78p/20190825T202139_press
 %
 %   Loading this file will add struct called Data to the Matlab workspace
 %
-%   load('/Users/ryan/Projects/Shimming/Acdc/202004_Ismrm/data/acdc_78p/20190825T202139_pressureProbeRec.mat')
+%   load('~/Matlab/shimming/example/acdc_78p/20190825T202139_pressureProbeRec.mat')
 % 
 %   Required is a script that takes the Siemens PMU recording as input and returns a file/struct equivalent to this. 
 %
--- a/example/shimdemo.m
+++ b/example/shimdemo.m
@@ -86,6 +86,21 @@ Params.isRealtimeShimming = 1 ;
 % e.g.
 %   Tx-freq adjustment ; none of the 8 multi-coil channels ; all 8 Prisma shim terms:
 Params.activeStaticChannelsMask = [ true ; false(8, 1) ; true(8, 1) ] ;
+% Note: To know which element of this vector list corresponds to which gradient/shim element, each hardware "X" has a 
+% "ShimSpecs_X" class defined, with the subfield .Id.channelNames() that has channel info. Example:
+% 
+% >> Specs = ShimSpecs_IUGM_Prisma_fit;
+% >> Specs.Id.channelNames
+% ans =
+%   8×1 cell array
+%     {'X'  }
+%     {'Y'  }
+%     {'Z'  }
+%     {'A20'}
+%     {'A21'}
+%     {'B21'}
+%     {'A22'}
+%     {'B22'}
 
 Params.isRealtimeShimming = 1 ;
 % Define which shim terms are to be included in the real-time shim:
--- a/example/shimdemo.m
+++ b/example/shimdemo.m
@@ -70,7 +70,7 @@ Field = FieldEval.modelfield( Fields ) ;
 Params = [] ;
 Shims  = ShimOpt_Greg( Params, Field ) ;
 
-% define the target region for shimming with a binary mask (same size as fieldB0.nii)
+% define the target region for shimming with a binary mask (same size as the Matlab variable: Field.img)
 %
 % by default, it will be the intersection of Field.Hdr.MaskingImage (where
 % reliable field values exist) and Shims.getshimsupport() (where shim reference
--- a/example/shimdemo.m
+++ b/example/shimdemo.m
@@ -86,21 +86,19 @@ Params.isRealtimeShimming = 1 ;
 % e.g.
 %   Tx-freq adjustment ; none of the 8 multi-coil channels ; all 8 Prisma shim terms:
 Params.activeStaticChannelsMask = [ true ; false(8, 1) ; true(8, 1) ] ;
-% Note: To know which element of this vector list corresponds to which gradient/shim element, each hardware "X" has a 
-% "ShimSpecs_X" class defined, with the subfield .Id.channelNames() that has channel info. Example:
+% Note: To know which element of this vector list corresponds to which gradient/shim element, you can run:
 % 
-% >> Specs = ShimSpecs_IUGM_Prisma_fit;
-% >> Specs.Id.channelNames
+% >> Shims.System.Specs.Id.channelNames
 % ans =
 %   8×1 cell array
-%     {'X'  }
-%     {'Y'  }
-%     {'Z'  }
-%     {'A20'}
-%     {'A21'}
-%     {'B21'}
-%     {'A22'}
-%     {'B22'}
+%     {'Ch1'}
+%     {'Ch2'}
+%     {'Ch3'}
+%     {'Ch4'}
+%     {'Ch5'}
+%     {'Ch6'}
+%     {'Ch7'}
+%     {'Ch8'}
 
 Params.isRealtimeShimming = 1 ;
 % Define which shim terms are to be included in the real-time shim:
--- a/example/shimdemo.m
+++ b/example/shimdemo.m
@@ -84,23 +84,17 @@ Shims  = ShimOpt_Greg( Params, Field ) ;
 shimVoi = Shims.getshimsupport() & Shims.Aux.getshimsupport() & Shims.Field.Hdr.MaskingImage ;
 Shims.setshimvolumeofinterest( shimVoi ) ;
 
-% Define which shim terms are to be included in the static shim:
-% e.g.
-%   Tx-freq adjustment ; none of the 8 multi-coil channels ; all 8 Prisma shim terms:
-Params.activeStaticChannelsMask = [ true ; false(8, 1) ; true(8, 1) ] ;
-% Note: To know which element of this vector list corresponds to which gradient/shim element, you can run:
+% Define which terms are to be included in the static shim optimization (boolean vector):
+%   [TX_FREQ ; CUSTOM_COIL ; SIEMENS_COIL ]
+% Where:
+%   - TX_FREQ is the Tx frequency of the MR sytem (obtained via frequency adjustement routine)
+%   - CUSTOM_COIL: Vector comprising each shim element. The name of each element is stored in: Shims.System.Specs.Id.channelNames  
+%   - SIEMENS_COIL: Vector comprising the gradient+shim elements. The name of each element is stored in: Shims.Aux.System.Specs.Id.channelNames
 % 
-% >> Shims.System.Specs.Id.channelNames
-% ans =
-%   8×1 cell array
-%     {'Ch1'}
-%     {'Ch2'}
-%     {'Ch3'}
-%     {'Ch4'}
-%     {'Ch5'}
-%     {'Ch6'}
-%     {'Ch7'}
-%     {'Ch8'}
+% NOTE: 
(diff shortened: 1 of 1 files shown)


example/shimdemo.m
```diff
@@ -70,7 +70,7 @@
 Params = [] ;
 Shims  = ShimOpt_Greg( Params, Field ) ;
 
-% define the target region for shimming with a binary mask (same size as Field.img)
+% define the target region for shimming with a binary mask (same size as fieldB0.nii)
```
jcohenadad: @rtopfer need your confirmation on this one...

jcohenadad: ah! i just get it. I thought it was a physical img/hdr image (old nifti format), but you were actually referring to the matlab variable `Field`.
OK, i will clarify

jcohenadad:
@rtopfer what is the difference between `Shims.getshimsupport()` & `Shims.Aux.getshimsupport() ` [here](https://github.com/neuropoly/realtime_shimming/blob/dc68ccc6ed49c13ddfda0ab13661ad7c51155371/example/shimdemo.m#L81)? Maybe worth documenting

jcohenadad:
@rtopfer any reason for repeating this line?

https://github.com/neuropoly/realtime_shimming/blob/dc68ccc6ed49c13ddfda0ab13661ad7c51155371/example/shimdemo.m#L84

https://github.com/neuropoly/realtime_shimming/blob/dc68ccc6ed49c13ddfda0ab13661ad7c51155371/example/shimdemo.m#L90

jcohenadad:
@rtopfer maybe we should redirect to a definition of the order of each element of the vector? (e.g. [Gx, Gy, Gz, etc.])

https://github.com/neuropoly/realtime_shimming/blob/dc68ccc6ed49c13ddfda0ab13661ad7c51155371/example/shimdemo.m#L87

example/shimdemo.m
rtopfer: Indeed, Field.img is a variable (specifically, "img" is a property of the MaRdI class and its subclasses FieldEval, ShimOpt, etc. It's always the image data itself.)

rtopfer:
> @rtopfer any reason for repeating this line?
> 
> https://github.com/neuropoly/realtime_shimming/blob/dc68ccc6ed49c13ddfda0ab13661ad7c51155371/example/shimdemo.m#L84
> 
> https://github.com/neuropoly/realtime_shimming/blob/dc68ccc6ed49c13ddfda0ab13661ad7c51155371/example/shimdemo.m#L90

Nope, just a typo!

jcohenadad:
> > @rtopfer any reason for repeating this line?
> > https://github.com/neuropoly/realtime_shimming/blob/dc68ccc6ed49c13ddfda0ab13661ad7c51155371/example/shimdemo.m#L84
> > 
> > https://github.com/neuropoly/realtime_shimming/blob/dc68ccc6ed49c13ddfda0ab13661ad7c51155371/example/shimdemo.m#L90
> 
> Nope, just a typo!

in that case i suggest to move it just before this line:
https://github.com/neuropoly/realtime_shimming/blob/dc68ccc6ed49c13ddfda0ab13661ad7c51155371/example/shimdemo.m#L94

rtopfer:
> @rtopfer maybe we should redirect to a definition of the order of each element of the vector? (e.g. [Gx, Gy, Gz, etc.])
> 
> https://github.com/neuropoly/realtime_shimming/blob/dc68ccc6ed49c13ddfda0ab13661ad7c51155371/example/shimdemo.m#L87

There is a property to help keep track of the ordering: Every ShimOpt instance has as a subfield an instance of ShimSpecs, within which there is .Id.channelNames()

E.g.
_Specs = ShimSpecs_IUGM_Prisma_fit;_
displaying Specs.Id.channelNames
yields the Siemens term names:
    'X'
    'Y'
    'Z'
    'A20'
    'A21'
    'B21'
    'A22'
    'B22'

jcohenadad:
> > @rtopfer maybe we should redirect to a definition of the order of each element of the vector? (e.g. [Gx, Gy, Gz, etc.])
> > https://github.com/neuropoly/realtime_shimming/blob/dc68ccc6ed49c13ddfda0ab13661ad7c51155371/example/shimdemo.m#L87
> 
> There is a property to help keep track of the ordering: Every ShimOpt instance has as a subfield an instance of ShimSpecs, within which there is .Id.channelNames()
> 
> E.g.
> _Specs = ShimSpecs_IUGM_Prisma_fit;_
> displaying Specs.Id.channelNames
> yields the Siemens term names:
> 'X'
> 'Y'
> 'Z'
> 'A20'
> 'A21'
> 'B21'
> 'A22'
> 'B22'

ok, i'll add this info in the shimdemo file

rtopfer:
> @rtopfer what is the difference between `Shims.getshimsupport()` & `Shims.Aux.getshimsupport() ` [here](https://github.com/neuropoly/realtime_shimming/blob/dc68ccc6ed49c13ddfda0ab13661ad7c51155371/example/shimdemo.m#L81)? Maybe worth documenting

in instantiating _Shims = ShimOpt_Greg() ;_
Shims gets a 2nd "auxiliary" ShimOpt_IUGM_Prisma_fit() object under Shims.Aux (since we know we're using the ACDC Greg coil with the Prisma).
But, indeed, it's not obvious, and there might a better way to organize this conceptually...

jcohenadad:
> > > @rtopfer maybe we should redirect to a definition of the order of each element of the vector? (e.g. [Gx, Gy, Gz, etc.])
> > > https://github.com/neuropoly/realtime_shimming/blob/dc68ccc6ed49c13ddfda0ab13661ad7c51155371/example/shimdemo.m#L87
> > 
> > 
> > There is a property to help keep track of the ordering: Every ShimOpt instance has as a subfield an instance of ShimSpecs, within which there is .Id.channelNames()
> > E.g.
> > _Specs = ShimSpecs_IUGM_Prisma_fit;_
> > displaying Specs.Id.channelNames
> > yields the Siemens term names:
> > 'X'
> > 'Y'
> > 'Z'
> > 'A20'
> > 'A21'
> > 'B21'
> > 'A22'
> > 'B22'
> 
> ok, i'll add this info in the shimdemo file

@rtopfer I don't understand this sentence:

> Every ShimOpt instance has as a subfield an instance of ShimSpecs, within which there is .Id.channelNames()

It seems that each hardware simply has a ShimSpecs_X class defined (i.e., I don't see how ShimSpecs_X is an instance of a subfield of ShimOpt).

rtopfer:
> > > > @rtopfer maybe we should redirect to a definition of the order of each element of the vector? (e.g. [Gx, Gy, Gz, etc.])
> > > > https://github.com/neuropoly/realtime_shimming/blob/dc68ccc6ed49c13ddfda0ab13661ad7c51155371/example/shimdemo.m#L87
> > > 
> > > 
> > > There is a property to help keep track of the ordering: Every ShimOpt instance has as a subfield an instance of ShimSpecs, within which there is .Id.channelNames()
> > > E.g.
> > > _Specs = ShimSpecs_IUGM_Prisma_fit;_
> > > displaying Specs.Id.channelNames
> > > yields the Siemens term names:
> > > 'X'
> > > 'Y'
> > > 'Z'
> > > 'A20'
> > > 'A21'
> > > 'B21'
> > > 'A22'
> > > 'B22'
> > 
> > 
> > ok, i'll add this info in the shimdemo file
> 
> @rtopfer I don't understand this sentence:
> 
> > Every ShimOpt instance has as a subfield an instance of ShimSpecs, within which there is .Id.channelNames()
> 
> It seems that each hardware simply has a ShimSpecs_X class defined (i.e., I don't see how ShimSpecs_X is an instance of a subfield of ShimOpt).

I meant every ShimOpt_X instance: 
e.g.
_Shims = ShimOpt_Greg() ;_
_Shims_ then has the subfield .System.Specs (within which there is .Id.channelNames, in this case they are just 'Ch1', 'Ch2', ... 'Ch8')

rtopfer:
> > > > @rtopfer maybe we should redirect to a definition of the order of each element of the vector? (e.g. [Gx, Gy, Gz, etc.])
> > > > https://github.com/neuropoly/realtime_shimming/blob/dc68ccc6ed49c13ddfda0ab13661ad7c51155371/example/shimdemo.m#L87
> > > 
> > > 
> > > There is a property to help keep track of the ordering: Every ShimOpt instance has as a subfield an instance of ShimSpecs, within which there is .Id.channelNames()
> > > E.g.
> > > _Specs = ShimSpecs_IUGM_Prisma_fit;_
> > > displaying Specs.Id.channelNames
> > > yields the Siemens term names:
> > > 'X'
> > > 'Y'
> > > 'Z'
> > > 'A20'
> > > 'A21'
> > > 'B21'
> > > 'A22'
> > > 'B22'
> > 
> > 
> > ok, i'll add this info in the shimdemo file
> 
> @rtopfer I don't understand this sentence:
> 
> > Every ShimOpt instance has as a subfield an instance of ShimSpecs, within which there is .Id.channelNames()
> 
> It seems that each hardware simply has a ShimSpecs_X class defined (i.e., I don't see how ShimSpecs_X is an instance of a subfield of ShimOpt).

Actually, the easiest thing for keeping track of the channels is to look at the .txt file that is output after running _Shims.optimizeshimcurrents()_. E.g. if you drag and drop that (or the "x_Stats.txt" files) into Keynote, they're automatically organized as nice tables with the channel names displayed). The default is to create a new folder in the current directory each time _.optimizeshimcurrents()_ is called, but the user can also change the directory with the optional input _Params.mediaSaveDir_

jcohenadad:
> > @rtopfer what is the difference between `Shims.getshimsupport()` & `Shims.Aux.getshimsupport() ` [here](https://github.com/neuropoly/realtime_shimming/blob/dc68ccc6ed49c13ddfda0ab13661ad7c51155371/example/shimdemo.m#L81)? Maybe worth documenting
> 
> in instantiating _Shims = ShimOpt_Greg() ;_
> Shims gets a 2nd "auxiliary" ShimOpt_IUGM_Prisma_fit() object under Shims.Aux (since we know we're using the ACDC Greg coil with the Prisma).
> But, indeed, it's not obvious, and there might a better way to organize this conceptually...

I don't understand the purpose of duplicating `ShimOpt_IUGM_Prisma_fit()` inside `Shims.Aux`. Are there scenario where the content of `Shims.Aux` would be different from the content of `Shims`?

rtopfer:
> > > @rtopfer what is the difference between `Shims.getshimsupport()` & `Shims.Aux.getshimsupport() ` [here](https://github.com/neuropoly/realtime_shimming/blob/dc68ccc6ed49c13ddfda0ab13661ad7c51155371/example/shimdemo.m#L81)? Maybe worth documenting
> > 
> > 
> > in instantiating _Shims = ShimOpt_Greg() ;_
> > Shims gets a 2nd "auxiliary" ShimOpt_IUGM_Prisma_fit() object under Shims.Aux (since we know we're using the ACDC Greg coil with the Prisma).
> > But, indeed, it's not obvious, and there might a better way to organize this conceptually...
> 
> I don't understand the purpose of duplicating `ShimOpt_IUGM_Prisma_fit()` inside `Shims.Aux`. Are there scenario where the content of `Shims.Aux` would be different from the content of `Shims`?

Joint optimization! 

You can create an instance of the Prisma shims on their own, e.g. _Shims =ShimOpt_IUGM_Prisma_fit()_, in which case .Aux is simply empty. however, since I wrote everything originally for the multi-coil shims, the optimization method actually expects .Aux to at least be defined, though its shim terms can be excluded with Params.activeStaticChannelsMask and Params.activeDynamicChannelsMask... Will have to adapt the code a bit for it to work when .Aux is empty.

jcohenadad:
> > It seems that each hardware simply has a ShimSpecs_X class defined (i.e., I don't see how ShimSpecs_X is an instance of a subfield of ShimOpt).
> 
> Actually, the easiest thing for keeping track of the channels is to look at the .txt file that is output after running _Shims.optimizeshimcurrents()_. E.g. if you drag and drop that (or the "x_Stats.txt" files) into Keynote, they're automatically organized as nice tables with the channel names displayed). The default is to create a new folder in the current directory each time _.optimizeshimcurrents()_ is called, but the user can also change the directory with the optional input _Params.mediaSaveDir_

I can indeed see these outputs (very convenient), however I think it is more intuitive for users to be able to understand the construction of the vector __before__ running `Shims.optimizeshimcurrents()` (chicken-and-egg problem). However, the problem is that when displaying `Shims.System.Specs.Id.channelNames`, only the custom shim channels appear. So a subsequent question would be: should the vector **always** be constructed in the following manner: `[TX FREQ ; CUSTOM_COIL ; SIEMENS_COILS ]`?  If yes, do you agree we should document it that way?

jcohenadad:
> > I don't understand the purpose of duplicating `ShimOpt_IUGM_Prisma_fit()` inside `Shims.Aux`. Are there scenario where the content of `Shims.Aux` would be different from the content of `Shims`?
> 
> Joint optimization!
> 
> You can create an instance of the Prisma shims on their own, e.g. _Shims =ShimOpt_IUGM_Prisma_fit()_, in which case .Aux is simply empty. however, since I wrote everything originally for the multi-coil shims, the optimization method actually expects .Aux to at least be defined, though its shim terms can be excluded with Params.activeStaticChannelsMask and Params.activeDynamicChannelsMask... Will have to adapt the code a bit for it to work when .Aux is empty.

Ah! Brilliant! I get it now: the "Aux" corresponds to the Siemens gradient/shim. This partially answers [this comment](https://github.com/neuropoly/realtime_shimming/pull/26#issuecomment-528091632).

I'll try to document this important info it in the shimdemo file

rtopfer:
> > > It seems that each hardware simply has a ShimSpecs_X class defined (i.e., I don't see how ShimSpecs_X is an instance of a subfield of ShimOpt).
> > 
> > 
> > Actually, the easiest thing for keeping track of the channels is to look at the .txt file that is output after running _Shims.optimizeshimcurrents()_. E.g. if you drag and drop that (or the "x_Stats.txt" files) into Keynote, they're automatically organized as nice tables with the channel names displayed). The default is to create a new folder in the current directory each time _.optimizeshimcurrents()_ is called, but the user can also change the directory with the optional input _Params.mediaSaveDir_
> 
> I can indeed see these outputs (very convenient), however I think it is more intuitive for users to be able to understand the construction of the vector **before** running `Shims.optimizeshimcurrents()` (chicken-and-egg problem). However, the problem is that when displaying `Shims.System.Specs.Id.channelNames`, only the custom shim channels appear. So a subsequent question would be: should the vector **always** be constructed in the following manner: `[TX FREQ ; CUSTOM_COIL ; SIEMENS_COILS ]`? If yes, do you agree we should document it that way?

Ah, indeed. The order will be that way if the user instantiated _Shims = ShimOpt_Custom_Coil_ (in which case the Siemens_Coil becomes Shims.Aux). However I do think the current naming scheme with the .Aux property is confusing insofar as the custom coils are probably better described as 'auxiliary' equipment... 
In any case, we can add a method to ShimOpt that reveals the available channel names and thus the input order for the Params.activeXChannelsMask terms, e.g. ShimOpt.getactivechannels()

jcohenadad:
> Ah, indeed. The order will be that way if the user instantiated _Shims = ShimOpt_Custom_Coil_ (in which case the Siemens_Coil becomes Shims.Aux). However I do think the current naming scheme with the .Aux property is confusing insofar as the custom coils are probably better described as 'auxiliary' equipment...
> In any case, we can add a method to ShimOpt that reveals the available channel names and thus the input order for the Params.activeXChannelsMask terms, e.g. ShimOpt.getactivechannels()

@rtopfer clarified in https://github.com/neuropoly/realtime_shimming/pull/26/commits/b5204f940c6e0e848e792f949823735250638865, let me know if this is correct.

rtopfer (approved):
Documentation improved

PR #34, opened 2019-10-02
Integrating Siemens PMU recordings with ProbeTracking

rtopfer:
@astroeva I'm reviewing the [code](https://github.com/neuropoly/realtime_shimming/compare/eao/21-siemensPMU?expand=1#diff-4a498b2fef8a933465627f8d4949f41c) 

few questions:
1. Difference between _data_blk1_ and _data_blk2_ ? I guess the former precedes the image acquisition? If so, is it worth keeping?

2. Line 69: _t_off = find(data == 5003);  % System uses identifier 5003 as trigger OFF_
Is the comment accurate? Seems 5003 signals the end of the recording.

3. About the .resp example included in the google drive, what image data does this correspond to? I assumed it would be from your modified field mapping seq. with the initial trigger output, but the _t_on_ count is 16 as opposed to 1? Was this an EPI? or do these _t_on_ (data==5000) points not correspond to the external trigger output?


evaalonsoortiz:
1. data_blk1 is bracketed by 6002 and 5002: I believe it is part of the data

2. I am not sure, the IDEA course notes for VD11 (I think this is the version that they are for, although it is not actually stated anywhere in them) has 5000 and 6000 and trigger ON and OFF. Other sources (https://cfn.upenn.edu/aguirre/wiki/public:pulse-oximetry_during_fmri_scanning) state that it is 5000 and 5003, but again, with no mention of software version. I can change the code so that everything over 4095 is considered non-data. We do not need to find triggers for our purpose. 

3. It should correspond to a field map series, without the trigger modification.

rtopfer:
Ok, I might be able to ask Siemens about some of these things tomorrow. I would like to know what the "triggers" mean (new slice? new volume/measurement?)

Your code has the resp. sampling rate at 400 Hz but the upenn link says "The third [value in the .resp file] is related to the “ticktime”, which is the time (in msec) between subsequent measurements, although the exact relationship is still being determined." If the link's description is accurate, then the sampling rate would be 50 Hz based on the 3rd value in .resp = 20. Wondering where 400 Hz came from?

evaalonsoortiz:
The sampling rates comes from difference between the start and stop times, divided by the number of data points. I hadn't noticed that comment in the upenn link, but if you look here: https://gist.github.com/rtrhd/6172339  these people mention 2.5 ms sampling as well.

I think that trigger refers to a trigger signal you can have the sequence output, such as the one I implemented in the same sequence. We can try running that and checking the PMU to see if there is a trigger code 5000. I imagine they're often used to align fMRI data with PMU signals.

rtopfer:
Looking at the _01 and _02.resp logs for acdc_72 more closely, I have no clue what the "5000" values indicate: the corresponding "realtime_fieldmap" GRE acquisitions have identical parameters, yet the _01.resp file has 32 instances of these "triggers", and _02.resp only has 11 !

*Edit*: Nevermind : Looks like the 5000's serve as indicators of periodicity, they seem to be inserted when the resp cycle repeats 
egRespTriggers

evaalonsoortiz:
Weird! Maybe we should ask the IDEA forum community?

rtopfer:
I think the "triggers" are there for gating certain sequences -- probably not immediately useful to us, so safe to ignore them as you suggested.  The upenn link you shared describes how to align the PMU logs with the corresponding images which I've implemented in _MaRdI.associateaux()_.
Briefly, user creates a ProbeTracking object using the PMU.resp file: 
_Probe = ProbeTracking( filename )_
loads the associated DICOMs as a MaRdI object:
_Phase = MaRdI( dicomDirectory )_
then makes a call to _Phase.associateaux( Probe )_ to crop+interpolate the PMU recording to the acquisition times (the interpolated Probe object is returned to _Phase.Aux_)

PR #52, opened 2020-03-10
Refactored the class model for image handling

rtopfer:
This pull request is to improve the class model for image handling: avoid the MaRdI "God" class! and make the code more modular.


rtopfer:
This refactor is too extensive for a single PR and will be handled iteratively. Closing PR.

Issue #60, opened 2020-04-08
Create a function to write nifti file


jcohenadad:
Currently we are writing nifti by going through dicom format, which is problematic because of re-normalization.

Suggestion: write a function that would write nifti directly, with json that woudl include useful metadata (freq, shim coeff, etc.)

jcohenadad:
Maybe look into official matlab functions:
- https://www.mathworks.com/help/images/ref/niftiread.html
- https://www.mathworks.com/help/images/ref/niftiwrite.html

jaystock:
How does the NeuroPoly team handle the 90 deg rotation and R-L flip that occurs in DICOM-to-NIFTI conversion?  Do you manually correct for this or is there a systematic way to keep track of the re-ordering of Matlab matrix dimensions?  I find that it's easy to accidentally reverse R and L.

jcohenadad:
> How does the NeuroPoly team handle the 90 deg rotation and R-L flip that occurs in DICOM-to-NIFTI conversion? Do you manually correct for this or is there a systematic way to keep track of the re-ordering of Matlab matrix dimensions? I find that it's easy to accidentally reverse R and L.

We deal with it by systematically relying on the nifti header when opening a data matrix. One thing that could be done, to simplify code, is to reorient the data in a specific orientation (eg RPI) before entering the optimization pipeline. This is indeed an important aspect that needs to be clarified in the code base.

PR #62, opened 2020-05-06
Fix the MaRdI.write function to output nifti correctly

jcohenadad:
This pull request fixes the MaRdI.write method to output nifti files correctly. It fixes issue #60

### DIFF
--- a/misc/MaRdI.m
+++ b/misc/MaRdI.m
@@ -1891,15 +1891,15 @@ Hdr.InPlanePhaseEncodingDirection  = Img.Hdr.InPlanePhaseEncodingDirection ;
 
 nSlices       = size( Img.img, 3 ) ;
 nEchoes       = numel( Img.getechotime() ) ;
-nAcquisitions = numel( Img.getacquisitiontime ) ;
+nAcquisitions = size( Img.getacquisitiontime , 3 ) ;
 
-nImg = nSlices*nEchoes*nAcquisitions ;
+nImg = nSlices*nEchoes*nAcquisitions ; 
 
 for iSlice = 1 : nSlices 
     for iEcho  = 1 : nEchoes 
         for iAcq = 1 : nAcquisitions  
 
-            iImg = iSlice*iEcho*iAcq ;
+            iImg = (iSlice - 1) * (nEchoes + nAcquisitions) + (iEcho - 1) * nAcquisitions + iAcq;
 
             disp( [num2str(100*iImg/nImg) '% ...'] ) ;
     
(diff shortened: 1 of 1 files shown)


po09i:
There was a problem when writing the nifti image. The MaRdI property stores the image data as double. The dicomwrite function assumes the data to be between 0 and 1 if the type is double so all values above 1 were changed to the maximum value of uint16. dicomwrite also accepts uint16 and type casting the image as uint16 resolves the issue for uint16 input images. [Fixed in this commit](https://github.com/shimming-toolbox/shimming-toolbox/pull/62/commits/863daceac27a0a0863db663f5c35b8aa37d7eaf0)

However, I assume that not all input images going into the MaRdI.write will have a uint16 types. (phase images for example)

jcohenadad:
@po09i wasn't the purpose of this PR to bypass matlab2dcm conversion and to implement direct matlab2nii conversion?

rtopfer:
indeed, I would suggest forgetting about `MaRdI.write` entirely: just add a wrapper method to `MaRdI` called `niftiwrite` which calls Matlab's [`niftiwrite`](https://www.mathworks.com/help/images/ref/niftiwrite.html)—most of the arguments (e.g. voxel size, slice orientation etc.) can probably be assigned automatically based on the `MaRdI.Hdr` values.  

i.e. A user with a `MaRdI`-type object `obj` should be able to call `niftiwrite( obj, ... )` and, ideally, be unaware that the call is actually mediated by the `MaRdI.niftiwrite()` wrapper method: the interface should appear identical to the Matlab function itself.

po09i:
> wasn't the purpose of this PR to bypass matlab2dcm conversion and to implement direct matlab2nii conversion?

Bypassing dcm2niix was a proposed solution. However, the following problem is originally a problem when writing dicoms.

> The dicomwrite function assumes the data to be between 0 and 1 if the type is double so all values above 1 were changed to the maximum value of uint16.

By fixing this problem, it also fixes the nifti saturation problem when using dcm2niix. 

If we go the niftiwrite wrapper approach, it also means that a dicomwrite wrapper method is required instead of the current write method. I see this solution as cleaner but writing a niftiwrite wrapper is effectively like doing dcm2niix's job since MaRdI is effectively a dicom in a Matlab structure. 

The other solution is writing the dicom properly (which has to be done regardless) and see if dcm2niix satisfies our expectations. I'm happy to go either way.

rtopfer:
>If we go the niftiwrite wrapper approach, it also means that a dicomwrite wrapper method is required instead of the current write method.

To be sure, `MaRdI.write` was meant to be precisely that—a wrapper to `dicomwrite`. The nifti functionality got tacked on in a hurry for a specific project, and at a time when Matlab's image processing toolbox didn't have built-in support for nifti (hence dcm2niix).

I think it would be better to get rid of the dcm2niix dependency here and, assuming it works just as well, use niftiwrite instead.

PR #67, opened 2020-05-08
Add CI and testing suite

jcohenadad:
The purpose of this PR is to:
- establish a continuous integration (CI) framework using Github actions;
- establish a test suite with [MOxUnit](https://github.com/MOxUnit/MOxUnit);
- upload mock data on newly-created [OSF repository](https://osf.io/ntb32/)

## Local testing

Below is a docker solution to mimik the online GH environment

<details>

```bash
docker pull ubuntu:16.04
docker run -it ubuntu:latest
# while inside docker
apt-get -y -qq update
apt-get -y install wget unzip emacs git
apt-get -y install octave liboctave-dev octave-pkg-dev
# install testing suite
git clone https://github.com/MOxUnit/MOxUnit.git
cd MOxUnit/
make install
# install repos
git clone https://github.com/shimming-toolbox/shimming-toolbox.git
cd shimming-toolbox/
git checkout -b jca/ci-github-action origin/jca/ci-github-action
# download mock data
wget -O data.zip https://osf.io/7d2j5/download
unzip data.zip 
```

</details>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,30 @@
+# Continuous integration workflow for shimming-toolbox
+
+name: CI
+
+on: [push]
+# on:
+#   push:
+#     branches: [ master ]
+#   pull_request:
+#     branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Download mock data
+      run: |
+        cd tests
+        wget -O data.zip https://osf.io/7d2j5/download
+        unzip data.zip
+        rm data.zip
+        cd ..
+
+    - name: Run MOxUnit tests
+      uses: joergbrech/moxunit-action@master
+      with:
+        tests: tests
+        src: misc Aux Coils external Img
--- a/tests/test_dummy.m
+++ b/tests/test_dummy.m
@@ -1,0 +1,25 @@
+function test_suite=test_dummy
+    try % assignment of 'localfunctions' is necessary in Matlab >= 2016
+        test_functions=localfunctions();
+    catch % no problem; early Matlab versions can use initTestSuite fine
+    end
+    initTestSuite;
+
+function test_abs_scalar
+    assertTrue(abs(-1)==1)
+    assertEqual(abs(-NaN),NaN);
+    assertEqual(abs(-Inf),Inf);
+    assertEqual(abs(0),0)
+    assertElementsAlmostEqual(abs(-1e-13),0)
+
+function test_abs_vector
+    assertEqual(abs([-1 1 -3]),[1 1 3]);
+
+function test_abs_exceptions
+    % GNU Octave and Matlab use different error identifiers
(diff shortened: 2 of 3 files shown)


kousu:
Wow https://github.com/joergbrech/moxunit-action/ is great! Unfortunately there's a lot of compatibility problems with the current code base and Octave that will need to be addressed first before we can run it in CI. You can see that Action uses Octave here

https://github.com/joergbrech/moxunit-action/blob/f7682ab17e6b552d39059d5d5c68fd2d176e67fc/entrypoint.sh#L113

and it's never going to not because Matlab needs a license to run which Github isn't going to pay for just for us.

So I am very much in support of going in this direction but we probably need to do a separate Octave-compatibility PR for it. Perhaps this PR could just add the test data -- and make sure the files in example/ work reproducibly on Matlab, the next PR could fix Octave compatibility, and the last one could integrate CI.

kousu:
Also I would recommend rebase-squashing all the commits to `.github/workflows/main.yml` together. There's no way to debug CI without uploading to CI, but once it's working there's nothing to be learned from keeping the experimental stage around.

jcohenadad:
> Also I would recommend rebase-squashing all the commits to .github/workflows/main.yml together. There's no way to debug CI without uploading to CI, but once it's working there's nothing to be learned from keeping the experimental stage around.

i know it's ugly... unfortunately i have some other emergencies to take care off and i won,t have the time to do a proper cleaning

po09i:
I squashed the commits to the best of my ability in this [branch](https://github.com/shimming-toolbox/shimming-toolbox/tree/ad/ci-github-actions-squashed). If this work, I can apply the changes to this branch.

jcohenadad:
@po09i could you do the squash directly in my branch and then leave your comments— thanks

.github/workflows/main.yml
```diff
@@ -0,0 +1,30 @@
...
+        cd tests
+        wget -O data.zip https://osf.io/7d2j5/download
+        unzip data.zip
+        rm data.zip
+        cd ..
+
+    - name: Run MOxUnit tests
+      uses: joergbrech/moxunit-action@master
+      with:
+        test: tests
```
po09i, line 29: The argument name should be called "tests" according to [mox-unit-action repo](https://github.com/joergbrech/moxunit-action/tree/f7682ab17e6b552d39059d5d5c68fd2d176e67fc). This should remove the warning that was thrown when running the step: run MoxUnit tests.

po09i: The argument name should be called "tests" according to [mox-unit-action repo](https://github.com/joergbrech/moxunit-action/tree/f7682ab17e6b552d39059d5d5c68fd2d176e67fc). This should remove the warning that was thrown when running the step: run MoxUnit tests.

po09i (changes requested):

.github/workflows/main.yml
jcohenadad: good point! would you mind fixing it directly in the branch? thx

po09i (approved):

PR #74, opened 2020-05-17
Create niftiwrite method fixes #60

### LINKED ISSUE #60: Create a function to write nifti file

po09i:
The goal of this PR is to write nifti files from the Mardi object including metadata. A niftiwrite method is used as a wrapper to Matlab's niftiwrite function to write the nifti files.

### DIFF
--- a/Img/MaRdI.m
+++ b/Img/MaRdI.m
@@ -1951,7 +1951,74 @@ end
 
 end
 % =========================================================================
+function niftiwrite(Img, filename, varargin)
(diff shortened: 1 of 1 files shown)


po09i:
On hold for refactoring

PR #80, opened 2020-05-19
Added logfile for realtime_zshim

### LINKED ISSUE #79: Log relevant data processing information for realtime_zshim.m in an external file
Can use matlab's "diary" function.

evaalonsoortiz:
fixes #79

### DIFF
--- a/Img/FieldEval.m
+++ b/Img/FieldEval.m
@@ -935,8 +935,6 @@ if isa( Fields, 'FieldEval' )
     Field.Hdr.MaskingImage   = Field.getvaliditymask( Params.maxAbsField ) ;
     Field.Aux.Data.p         = pMean ; 
     
-    % scale RIRO by RMSE of physio signal
-    Riro.img( mask )        = pShiftRms .* x(nVoxelsVoi+1:end) ;
     Riro.img( ~mask )       = 0 ;
     Riro.Hdr.MaskingImage   = mask ; % not redundant: mask used in following call to .getvaliditymask()
     Riro.Hdr.MaskingImage   = Riro.getvaliditymask( Params.maxFieldDifference ) ;
--- a/Img/FieldEval.m
+++ b/Img/FieldEval.m
@@ -935,6 +935,8 @@ if isa( Fields, 'FieldEval' )
     Field.Hdr.MaskingImage   = Field.getvaliditymask( Params.maxAbsField ) ;
     Field.Aux.Data.p         = pMean ; 
     
+    % scale RIRO by RMSE of physio signal	
+    Riro.img( mask )        = pShiftRms .* x(nVoxelsVoi+1:end); 
     Riro.img( ~mask )       = 0 ;
     Riro.Hdr.MaskingImage   = mask ; % not redundant: mask used in following call to .getvaliditymask()
     Riro.Hdr.MaskingImage   = Riro.getvaliditymask( Params.maxFieldDifference ) ;
--- a/example/realtime_zshim.m
+++ b/example/realtime_zshim.m
@@ -81,10 +81,10 @@ if nargin > 1
     sortedDicomDir = input('Enter the desired path for the sorted images: ');
     if (varargin{1} == 1)
         % copy the files if optional boolean is 1
-        MaRdI.sortimages( unsortedDicomDir, sortedDicomDir, 1 );
+        sortdicoms( unsortedDicomDir, sortedDicomDir, 1 );
     elseif (varargin{1} == 0)
         % move the files is optional boolean is 0
-        MaRdI.sortimages( unsortedDicomDir, sortedDicomDir, 0 );
+        sortdicoms( unsortedDicomDir, sortedDicomDir, 0 );
     end
     % copy respiratory trace file from mounted drive to local directory
     unix('cp /SYNGO_TRANSFER/SYNGO_TRANSFER/PMUresp_signal.resp .')
--- a/example/realtime_zshim.m
+++ b/example/realtime_zshim.m
@@ -176,7 +176,7 @@ B0Fields.associateaux( Pmu );
 
 
 %% ------------------------------------------------------------------------
-% rescale and compute z-gradients  
+% compute z-gradients  
 %% ------------------------------------------------------------------------
 
 GzFields = B0Fields.copy();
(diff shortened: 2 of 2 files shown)


example/realtime_zshim.m
```diff
@@ -67,10 +81,10 @@ function realtime_zshim(scan_obj, varargin)
     sortedDicomDir = input('Enter the desired path for the sorted images: ');
     if (varargin{1} == 1)
         % copy the files if optional boolean is 1
-        sortdicoms( unsortedDicomDir, sortedDicomDir, 1 );
```
rtopfer: @evaalonsoortiz `sortdicoms` is now a [standalone function ](https://github.com/shimming-toolbox/shimming-toolbox/blob/master/misc/sortdicoms.m) on master. (merge current master into the issue branch before testing)

rtopfer (changes requested):
call needs to be to `sortdicoms` and not `MaRdI.sortdicoms`

Img/FieldEval.m
```diff
@@ -935,8 +935,6 @@
     Field.Hdr.MaskingImage   = Field.getvaliditymask( Params.maxAbsField ) ;
     Field.Aux.Data.p         = pMean ; 
     
-    % scale RIRO by RMSE of physio signal
```
rtopfer, line 938: @evaalonsoortiz  Does this need to be removed?
Normalizing by the RMSE actually makes Riro.img somewhat interpretable: i.e. if subject A's breathing causes more b0 variation than subject B, one can see that in a side-by-side plot of Riro.img for each subject.
Without rescaling, one can't make the comparison since Riro.img will be affected arbitrarily by the sensitivity of the bellows (i.e. if the belt is fastened more tightly to subject A than subject B—even if subject B actually produces more b0 variation while breathing—Riro.img for subject A will likely look a lot more intense).

example/realtime_zshim.m
evaalonsoortiz: done:
https://github.com/shimming-toolbox/shimming-toolbox/blob/eca9b6f9cc4641caaf0d6c0ff83222f74edd02c0/example/realtime_zshim.m#L84

Img/FieldEval.m
evaalonsoortiz: I think that scaling is useful, when you're plotting the images (the user can do this him/herself when plotting). However, I think it's logical that the output of a fitting method should return all the constants with the same units. Otherwise, it could be really confusing.

rtopfer: ah I think it's that the comment in the code is misleading—it's the multiplication by the RMSE that does exactly that^: converts Riro.img to _Hz_. Without the multiplication, Riro.img is in Hz/unit PMU.

I agree though, the handling of all RIRO-related stuff in the code is all overly confusing (e.g. in terms of units, and _where_ the the variable is accessed—buried as it is as Field.Model.Riro.img)...

I'm going to propose we abandon the whole MaRdI class structure (from which FieldEval inherits). If we go that way then the form of all this stuff^ will change (hopefully easier to tidy it up).

evaalonsoortiz: Right, I made a mistake in my previous comment. If you are requesting a linear fit (a + b*pressure) you expect that the units of b are "something"/[units of pressure]. If modelfield() returns anything else, it's easy to miss that since it is buried in another file.

example/realtime_zshim.m
```diff
@@ -51,10 +52,23 @@ function realtime_zshim(scan_obj, varargin)
 % Ryan Topfer, MSc. ryan.topfer@polymtl.ca
 % Eva Alonso Ortiz (EAO), PhD. eva.alonso.ortiz@gmail.com 
 %
-% LAST MODIFIED: Jan. 2020
+% LAST MODIFIED: May 2020
```
jcohenadad: remove the date-- that's what github is for

Img/FieldEval.m
rtopfer: Indeed. There are pros+cons to both outputs.

Anyway, the scaling factor is returned in `Field.Model.Riro.Aux.Data.p`. For now, can you just rescale in _realtime_zshim.m_ and leave `FieldEval.m` as it is? the original scaling is already assumed in other dependent parts of the code, e.g. `ShimOpt.m`, so they would have to be changed as well.

example/realtime_zshim.m
rtopfer: we can vote on it but I suggest from now on we do away with the tags as well, i.e.
```
% Ryan Topfer, MSc. ryan.topfer@polymtl.ca
% Eva Alonso Ortiz (EAO), PhD. eva.alonso.ortiz@gmail.com 
```

jcohenadad: also a useful feature of github: https://github.com/shimming-toolbox/shimming-toolbox/graphs/contributors

rtopfer: precisely. (+commit history already reveals who did what)

Img/FieldEval.m
evaalonsoortiz: Done

example/realtime_zshim.m
evaalonsoortiz: Done

evaalonsoortiz: > remove the date-- that's what github is for

Thinking about this further: Including the names and dates seems to be the habit that people generally have. I know that git does this for you, but it takes more digging and knowhow than simply glancing at the header. When I look at code, if I see a particular snippet that interests me, I find it useful to know right away who wrote it so I can potentially contact them for more info.

jcohenadad: Including the date has indeed been a habit when people were not using version control framework, but this habit has now evolved within modern software engineering framework. There are of course many exceptions to this rule, but I see strong arguments against maintaining the date, the strongest being that maintaining a date is prone to error, for example in [this file](https://github.com/shimming-toolbox/shimming-toolbox/blob/master/example/realtime_zshim.m#L54), which indicates "LAST MODIFIED: Jan. 2020" whereas last commit is from [15 days ago](https://github.com/shimming-toolbox/shimming-toolbox/commit/19b6d55b148bf0185738edd6d159806196958da3). There are other examples, and there will be more in the future. I'd rather have no date than a wrong date. 

As for the authors, I feel less strongly about it. However, two things to keep in mind:
- The purpose of an open-source project is that multiple people contribute to the codebase. So, an initial code could indeed be explicitly signed in the file header, however people should be aware that the file is reviewed by the team during PR (and code review *is* an intellectual contribution) and is prone to being modified by others as time goes by (bug fixes, new features, etc.). 
- The purpose of the "Support/Contact" link in the shimming-toolbox website is precisely to redirect users to the team. 

But again, I am not opposed to a 2-tear system, i.e., mention the "initial contribution" (e.g. Created by: XXX) and having a github-based contribution tracking system, which is referenced as a link in the "Contributors" section of the website.

jcohenadad (approved):
I have not tested the code (only glanced at it)

rtopfer (approved):

PR #81, opened 2020-05-19
Set up CI and doc build with Azure Pipelines

kousu:
Sets up CI that can run against Matlab. We already have an Azure account set up with a self-hosted runner tristano.neuro.polymtl.ca with Matlab installed for https://github.com/qMRLab/qMRLab/.

So, this is #67 but using Azure instead of Github (which is sort of silly since Microsoft built both of them).

It would be nice to add tristano as a [self-hosted Github runner](https://help.github.com/en/actions/hosting-your-own-runners/adding-self-hosted-runners) because Github Actions is a lot easier to work with from Github than Azure -- **but** for this very large all-caps caveat: 
> Forks of your public repository can potentially run dangerous code on your self-hosted runner machine by creating a pull request that executes the code in a workflow.

Azure prompts for permission to be granted when a pipeline is first run on a given runner. Github Actions apparently hasn't implemented this yet? This seems bizarre; why would forks retain access to private runners? That's got to be wrong, or at least, in the process of becoming wrong. I don't want to be maintaining two CI infrastructures in parallel.

It might be useful to run on both Matlab and Octave, with some tests that depend on Matlab allowed to soft-fail on Octave (for comparison: pytest calls these [xfails](https://docs.pytest.org/en/latest/skipping.html)), since Octave can be parallelized easier and so will give faster feedback, but I don't really know yet; WIP WIP WIP.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -14,5 +14,12 @@ steps:
     # And worse, our Matlab license only allows one process at a time (TODO: verify this),
     # so a forgotten `quit` will block all tests on all branches.
     # https://www.mathworks.com/matlabcentral/answers/523194-matlab-script-in-batch-from-unix-command-line
-    matlab -nojvm -nodisplay -nosplash -r "disp('Hello from Matlab'); quit"
+    matlab -nodisplay -nosplash -r "disp('Unit tests!'); quit"
   displayName: 'Unit Tests'
+- script: |
+    # Beware: you MUST put `quit` at the end; Matlab does not auto-quit.
+    # And worse, our Matlab license only allows one process at a time (TODO: verify this),
+    # so a forgotten `quit` will block all tests on all branches.
+    # https://www.mathworks.com/matlabcentral/answers/523194-matlab-script-in-batch-from-unix-command-line
+    matlab -nodisplay -nosplash -r "disp('Build the docs!'); quit"
+  displayName: 'Build doc'
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -9,10 +9,10 @@
 pool: 'Tristano'
 
 steps:
-- displayName: 'Unit Tests'
-  script: |
+- script: |
     # Beware: you MUST put `quit` at the end; Matlab does not auto-quit.
     # And worse, our Matlab license only allows one process at a time (TODO: verify this),
     # so a forgotten `quit` will block all tests on all branches.
     # https://www.mathworks.com/matlabcentral/answers/523194-matlab-script-in-batch-from-unix-command-line
     matlab -nojvm -nodisplay -nosplash -r "disp('Hello from Matlab'); quit"
+  displayName: 'Unit Tests'
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -21,5 +21,5 @@ steps:
     # And worse, our Matlab license only allows one process at a time (TODO: verify this),
     # so a forgotten `quit` will block all tests on all branches.
     # https://www.mathworks.com/matlabcentral/answers/523194-matlab-script-in-batch-from-unix-command-line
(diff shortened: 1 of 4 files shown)


kousu:
Aha: https://github.community/t5/GitHub-Actions/Self-hosted-runner-security-with-public-repositories/m-p/50172#M7723

We can use `tristano` with Github Actions, but we have to make sure to *not* run them on PRs (the same way I [disabled `travis-ci/pr`](https://github.com/neuropoly/spinalcordtoolbox/issues/2674)). PRs could run Octave on Github's infra, but `tristano` + matlab should be reserve to branches in the repo.

This does seem kind of fragile. It would be easy to open a cybersecurity risk since it fails-open instead of fails-closed.

jcohenadad:
@kousu that sounds like a reasonable approach. We can add a big warning in the "contribution" and/or wiki to explain why github actions should be disabled for PRs (until this issue is resolved-- if it is one day). So, does it mean that we won't need to rely on Azure pipeline anymore? I like the idea of not having to learn another CI environment. It also means that we can choose a more performant station (although this bit could be reserved for later, if we find tristano is too slow).

azure-pipelines.yml
```diff
@@ -0,0 +1,18 @@
...
+#- master
+
+# use the self-hosted runner at tristano.neuro.polymtl.ca
+# this has Matlab installed and a license to use it.
+pool: 'Default'
+
+steps:
+- script: |
+    # Beware: you MUST put `quit` at the end; Matlab does not auto-quit.
+    # And worse, our Matlab license only allows one process at a time (TODO: verify this),
```
agahkarakuzu, line 14: With the license we have on Tristano, concurrent runs are possible, the issue is if Azure build pipeline allows such multi-threading, as collecting unit test artifacts will need to be parallelized. If possible, we should better set`maxParallel: 2` with this machine. 

[Here](https://docs.microsoft.com/en-us/azure/devops/pipelines/licensing/concurrent-jobs?view=azure-devops#self-hosted-cicd) you can find details on parallelism.

azure-pipelines.yml
```diff
@@ -0,0 +1,18 @@
+# Run shimming-toolbox unit tests.
+
+# TODO:
+#trigger:
+#- master
+
+# use the self-hosted runner at tristano.neuro.polymtl.ca
+# this has Matlab installed and a license to use it.
+pool: 'Default'
```
agahkarakuzu, line 9: If you have non-matlab jobs, you can use a Microsoft hosted agent in parallel (you can use up to 10) to remove some burden off the shoulders of little tristano :)

agahkarakuzu:
🎊 badge for this branch:
[Build Status](https://dev.azure.com/neuropoly/RTShimming/_build/latest?definitionId=10&branchName=azure-pipelines)

Badge for master (should look nice on README):

[Build Status](https://dev.azure.com/neuropoly/RTShimming/_build/latest?definitionId=10&branchName=master)

agahkarakuzu:
> We can use tristano with Github Actions

@kousu @jcohenadad  FYI, GitHub actions  self-hosted runners do not have a good solution at an organization level yet. 

On Azure, you can associate certain actions with approved users, that big warning on GitHub self hosted runners + [WIP] for organization repos were repelled me from giving them a try for the time being. 

Equivalent of that warning on Azure:

https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#validate-contributions-from-forks

agahkarakuzu:
You should definitely check this checkbox:

image

On each PR, someone from the team can check files changed and confirm. I turned off fork builds for good measure anyway.

jcohenadad:
@kousu i've extended the scope of this PR to also include doc build, hope that's ok

kousu:
I think the doc build part should be isolated to https://github.com/shimming-toolbox/shimming-toolbox.org/pull/31 *or* we should merge these two repos.

PR #86, opened 2020-05-22
BUG: Fix purported Windows compatibility issue: Rename  subdir to

### LINKED ISSUE #85: [BUG] Impossible to clone the repository in Windows
Environment: Windows
Step to reproduce: git clone the project via https
Expected results: successful clone
Actual results: failed clone
Data that caused the issue: It seems that the 'aux' folder is a Windows reserved filenames and makes it impossible to create the corresponding folder in the git local repository. I think we should make sure that every filename in the project is compatible with the windows environment.

Information about Windows valid names https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file

image

rtopfer:
Rename `Aux` subdirectory to `aux_hardware`. Fixes #85


jcohenadad:
How does renaming Aux subdirectory to Aux_hardware fix the issue? Some context needed here

gaspardcereza:
@jcohenadad the 'aux' filename is reserved by windows so during the git clone, git cannot create a file named 'aux'. As aux_hardware isn't reserved, it should indeed fix the issue.

rtopfer:
i can't test it right now (hoping someone else has Windows?) but this is what was claimed:
https://github.com/shimming-toolbox/shimming-toolbox/issues/85#issue-623215205

—supported by this discussion:
https://stackoverflow.com/questions/36225708/cannot-create-a-directory-named-aux-or-starting-by-aux-on-windows-8-1

jcohenadad:
I see, Windows OS prevents from creating a directory called "Aux" at any place in the disk. This is crazy.

jcohenadad:
it seems like the folder was renamed "aux_hardware" (328faf1938a5fa93f038f15af40106304584c888) and not "Aux_hardware" as claimed [here](https://github.com/shimming-toolbox/shimming-toolbox/pull/86#issue-421944678). This discrepancy will create problems in Unix (where case matters).

rtopfer:
> it seems like the folder was renamed "aux_hardware" ([328faf1](https://github.com/shimming-toolbox/shimming-toolbox/commit/328faf1938a5fa93f038f15af40106304584c888)) and not "Aux_hardware" 

changed commit message

jcohenadad:
> > it seems like the folder was renamed "aux_hardware" ([328faf1](https://github.com/shimming-toolbox/shimming-toolbox/commit/328faf1938a5fa93f038f15af40106304584c888)) and not "Aux_hardware"
> 
> changed commit message

it's not only the commit message-- the folder was actually renamed "aux_hardware", see: https://github.com/shimming-toolbox/shimming-toolbox/tree/214035f5d0c4a0e492c1708400aa7175ac75cf9d

EDIT: regardless, we should also discuss convention for file name (in this case, it would be a mix snake_case and PascalCase). I am usually in favor of small caps for folder names, because it is easier to navigate within the terminal (no need to press shift). But, would that cause confusion with the class-based names (Img+, etc.)

EDIT: the convention for folder name is now: snake_case. The body of the PR has been updated accordingly.

po09i (approved):

jcohenadad (changes requested):
comment: https://github.com/shimming-toolbox/shimming-toolbox/pull/86#issuecomment-632740925

jcohenadad (approved):

PR #92, opened 2020-05-26
Rt/issue75 shim specs json

### LINKED ISSUE #75: REF: ShimSpecs—replace child classes with parameter files (.json)
The family of classes ([e.g.](https://github.com/shimming-toolbox/shimming-toolbox/blob/master/Coils/Shim_Greg/ShimSpecs_Greg.m)) consists exclusively of properties (shim system specifications).  

Rather than defining Matlab child classes for each new shim system, these attributes should just be listed in .json files—then used to initialize the properties of a single [ShimSpecs](https://github.com/shimming-toolbox/shimming-toolbox/blob/master/Coils/Shim_Template/ShimSpecs.m) class (which can also validate them, e.g. check number of channels is a positive integer, etc.).

rtopfer:
This PR serves to close #75 — replacing `ShimSpecs` subclasses with .json files.

This is an initial step toward refactoring the way shim systems are currently represented in the code base: Once completed, a minor refactor of the currently abstract `ShimOpt` class should allow us to dispense with several (if not all) of the `ShimOpt` child classes, thereby simplifying the code base and making it very easy to define new shim systems as only the calibration maps and .json file should then be required!


+img/read_nii.m
```diff
@@ -0,0 +1,37 @@
+function [img,info,json] = read_nii( niiFile )
+%READ_NII Load NIfTI image and header; and, when present, the accompanying .json sidecar
```
jcohenadad: not sure if it's of any help, but for the records here is how SPM deals with BIDS: https://en.wikibooks.org/wiki/SPM/BIDS

Coils/Shim_Greg/ShimOpt_Greg.m
```diff
@@ -24,11 +24,16 @@
 Shim.Field = [] ;       
 Shim.Model = [] ;
 Shim.Aux   = [] ;
-Shim.System.Specs    = ShimSpecs_Greg() ; 
-Shim.System.currents = zeros( Shim.System.Specs.Amp.nActiveChannels, 1 ) ; 
 
+[ Field, Params, Specs] = ShimOpt.parseinput( varargin ) ;
+
+if isempty(Specs)
+    Shim.System.Specs = ShimSpecs_Greg() ;
```
jcohenadad, line 22: didn't we agree to use camelCase for variables? so having something like this:
```suggestion
    shim.System.Specs = ShimSpecs_Greg() ; 
```

or maybe that would break the code somewhere else and we should do another specific PR for changing formatting?

rtopfer: baby steps!
We'll get rid of `ShimOpt_Greg.m` in a separate PR and change the case in the remaining classes.

jaystock:
For SMS acquisitions, would it make sense to explicitly store the slice ordering and slice group information in the .json file?  This information is in the twix meas.dat header, but I think it's also buried in the DICOM header, though it might not be parsed by normal DICOM header header scripts in Matlab.

jcohenadad:
> For SMS acquisitions, would it make sense to explicitly store the slice ordering and slice group information in the .json file? This information is in the twix meas.dat header, but I think it's also buried in the DICOM header, though it might not be parsed by normal DICOM header header scripts in Matlab.

Excellent idea. @rtopfer or @po09i can you look if already done by dcm2niix and if not, open an issue on their repos and document here via new issue. Thx

+b0shim/+coils/+greg/Contents.m
```diff
@@ -0,0 +1,8 @@
+%b0shim.coil.greg 8-channel AC/DC 3T neck coil
```
jcohenadad, line 1: to look nice on matlab's doc:
```suggestion
%% b0shim.coil.greg
% 8-channel AC/DC 3T neck coil
```
i suggest applying this format to the other Contents.m files. Also see https://github.com/shimming-toolbox/shimming-toolbox/pull/134#issuecomment-646825888

jaystock: Not sure if this is the most appropriate thread to post this, but I found a very fast way to interpolate the shim coil profile maps into the grid used by the target volume (for example the EPI volume) to prepare the coil field maps for shim optimization calculations.  I'm not sure how Ryan handled this in the original code.

`unix('flirt -in coil_field_maps.nii -ref epi_target.nii -applyxfm -usesqform -out coil_field_maps_interp')`

This seems to be ~100x faster than my previous approach which used scatteredInterpolant in Matlab.  Matlab is extremely slow if you use "scattered" coordinates for the interpolation rather than monotonically increasing points such as those generated by meshgrid.  The trick is to use the -usesqform flag in FLIRT, so that it will just apply a coordinate transform/interpolation without rotating or moving the image to register them.  

However, I'm not sure if we want to commit to relying on FSL in the GUI.   We might want to implement our own fast interpolator using the same methods as FLIRT.

jcohenadad: Thank you @jaystock, I moved the discussion to a specific issue thread https://github.com/shimming-toolbox/shimming-toolbox/issues/159

Issue #91, opened 2020-05-25
`niftiread()`: rescaling phase in Matlab to [-pi,pi]


rtopfer:
A function is needed to shift+rescale phase images and recast to double from int16,
i.e. to perform what this bit of code in [MaRdI](https://github.com/shimming-toolbox/shimming-toolbox/blob/16dae3169400d7a6e573ce80fde1bd2fc4dfe296/Img/MaRdI.m#L2145) does for DICOM's

- `info = niftiinfo( phase.nii ) ;` : the `info` struct generally contains `.MultiplicativeScaling` and `.AdditiveOffset` fields which are probably needed

+ question: Will the conversion formula be consistent across sites/sequences? The last paragraph in [_Processing Steps - Step 2_](https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/FUGUE/Guide_=) on the FSL/FUGUE page suggests it would not; in which case, we might want a site-specific configuration file for field mapping.

- btw: @evaalonsoortiz i noticed acquisitions 09-10 from ACDC108p (your custom sequence) don't have the normal rescaling info in the phase dicom header (so, the converted nifti's don't have it either). I'm not sure if there is a way to consistently determine what the actual phase (in radians) is without that. i.e. if there are wraps, a priori: 0=-pi, and 4095=+pi; but if there aren't any wraps present (e.g. max raw phase = 2000) then i think it might be impossible to determine the correct range.

evaalonsoortiz:
@rtopfer Hmm ... strange that's missing. What is the header name? Are you sure the range varies per sequence? I thought it only varied according to the software version and/or vendor.

rtopfer:
RescaleIntercept and RescaleSlope—tags: (0028,1052) and (0028,1053).

No idea whether it actually varies between sequences: I think it's been consistent across the few Siemens GRE sequences we've used but the disclaimer on the FSL/FUGUE page is disconcerting since obviously we want an automated solution for the rescaling and ideally that wouldn't require a massive if/else tree...

PR #97, opened 2020-06-02
Rt/read nii

rtopfer:
This PR: 
- Introduces a `readnifti` wrapper function _read_nii.m_ inside a new package `+imutils` 
+ Auto-scaling functionality for Siemens phase images is performed by default when calling `read_nii` and the json sidecar is available—addressing issue #91 (at least for the specific Siemens case)


+imutils/read_nii.m
```diff
@@ -0,0 +1,189 @@
...
+    [img, info] = rescale( img, info );
+end
+
+end
+
+% -----------------------------------------------------------------------------
+%% Local functions
+% -----------------------------------------------------------------------------
+function [img, info] = convert_siemens_phase( img0, info0 )
+%CONVERT_SIEMENS_PHASE Return phase img in rad (recast from int to single)
```
jcohenadad: I would add a comment about the assumption that the input image pixel intensity is scaled between 0 and 4095, and that corresponds to [-pi,pi[

jcohenadad (approved):

PR #119, opened 2020-06-16
Establish a workflow for processing niftis

po09i:
The purpose of this PR is to establish a workflow for processing niftis.

dicom_to_nifti : Takes unsorted dicoms and processes them using dcm2bids
load_niftis : Uses the bids folder structure to load niftis in the desired format.

The experiment uses those functions to show the workflow.


po09i:
Notes : 
- What should be done of the README inside the videotape folder that has a lot of info for octave compatibility issues?
- Should I squash rebase into a single commit?

jcohenadad (approved):
given we are pressed with time and many "brains" have already worked on that branch, i consider it safe to merge

po09i:
What about the README inside the videotape folder @jcohenadad ?

jcohenadad:
> What about the README inside the videotape folder @jcohenadad ?

ah... not good (oversight-- sorry). So:
- get rid of the README
- the chunk about the dependency/procedure goes in the experiment.m 
- the chunk about octave compatibility goes under the wiki (create a new entry "Octave")

that's the best i can think of for now

PR #125, opened 2020-06-16
B0 mapping based on LLSf

### LINKED ISSUE #117: Implement the Linear Least Square fit algorithm (LLSf)
The magnetic (B0) field can be measured from a series of gradient echo (GRE) images through a linear least squares regression of the phase of the GRE images vs. the echo times (TEs) [1]. Phase wraps can be removed by adding or subtracting 2π whenever the phase across successive TEs differs by more than π. 

LLSf-based B0 mapping has been implemented [here](https://github.com/evaalonsoortiz/LLSf_B0mapping). This code can be used as a starting point for integrating this approach within the shimming-toolbox. 

[1] Reber, P.J., et al., Correction of off resonance-related distortion in echo-planar imaging using EPI-based field maps. Magn Reson Med, 1998. 39(2): p. 328-30.

evaalonsoortiz:


evaalonsoortiz:
I think this PR is ready for review/feedback.

Both field maps shown below are generated using high-resolution multi-echo (4 echoes) gradient echo data obtained at 7T, which can be found here: https://mfr.ca-1.osf.io/render?url=https://osf.io/s8qhk/?direct%26mode=render%26action=download%26mode=render

B0 map obtained from a linear fit of the phase evolution vs. time:
image

B0 map obtained using a the 4-quadrant arctangent of the first 2 echoes:
image

The windowing is the same for both field maps. No spatial unwrapping was used. Temporal unwrapping was used in the first case, but the smallest deltaTE is 2.7 ms so multiple phase wraps per echo spacing is expected in high deltaB0 regions.

kousu:
I'm having trouble reproducing your example. So far I've put together what I know (and what you told me in the chatroom) into this script:

```
$ cat example/llsf.m 
% this should be run from inside the examples/ folder
addpath(genpath('..'));


% Relative path for the data
data = '7T_fieldmaps/';

% --------------------------------------
%% Download data when not already present
if ~isfolder( data )
    url = 'https://osf.io/s8qhk/download?version=1' ;
    fprintf( ['\n Downloading test data...\n URL=' url '\n'] ) ;
    unzip(url, '.') ;
end

% convert to nifti
% TODO: use imutils.dicom_to_nifti
!mkdir 7T_fieldmaps/ASPIRE_P_A_GRE_JASON_ASPIRE_100SLI_2MMISO_0006/nifti
!dcm2niix -o 7T_fieldmaps/ASPIRE_P_A_GRE_JASON_ASPIRE_100SLI_2MMISO_0006/nifti 7T_fieldmaps/ASPIRE_P_A_GRE_JASON_ASPIRE_100SLI_2MMISO_0006/
!mkdir 7T_fieldmaps/ASPIRE_M_A_GRE_JASON_ASPIRE_100SLI_2MMISO_0007/nifti
!dcm2niix -o 7T_fieldmaps/ASPIRE_M_A_GRE_JASON_ASPIRE_100SLI_2MMISO_0007/nifti 7T_fieldmaps/ASPIRE_M_A_GRE_JASON_ASPIRE_100SLI_2MMISO_0007/

[mag_data, mag_info, mag_json] = imutils.load_niftis('7T_fieldmaps/ASPIRE_M_A_GRE_JASON_ASPIRE_100SLI_2MMISO_0007/nifti');
[ph_data, ph_info, ph_json] = imutils.load_niftis('7T_fieldmaps/ASPIRE_P_A_GRE_JASON_ASPIRE_100SLI_2MMISO_0006/nifti');

mag_data = double(squeeze(mag_data));
ph_data = double(squeeze(ph_data));

ph_data = rescalePhaseImage(ph_data);

compl_vol(:,:,:,:) = mag_data(:,:,:,:).*exp(1i*ph_data(:,:,:,:));

[delf, offset, STDX, MSE] = B0_multiecho_linfit(compl_vol, ph_json)
```

which I run like this (on a unix machine):

```
$ (cd example; matlab -r "run('llsf.m');")
```

This runs, and I'm dropped to a matlab shell with 

```
>> who

Your variables are:

MSE        STDX       compl_vol  data       delf       mag_data   mag_info   mag_json   offset     ph_data    ph_info    ph_json    
>> size(delf) 

ans =

   112   112   100

>> size(offset)

ans =

   112   112   100

>> size(STDX)

ans =

   112   112   100     2

>> size(MSE)

ans =

   112   112   100
```

I don't know what to do next though. How did you make those images? Did you have to do further processing?


This script could become a unit test, but before we do that I want to figure out how to treat the sample data as a [fixture](https://en.wikipedia.org/wiki/Test_fixture); I know @agahkarakuzu has advice on testing in https://github.com/agahkarakuzu/eda_organized, so maybe there's advice in there.

evaalonsoortiz:
@kousu You could have a look at delf (the frequency map) using `imagesc` in matlab or save it as a nifti using `niftiwrite(delf,'delf_map.nii')` and then view it using fsleyes. 

I will add an example script. One thing I wanted to consult with everyone about is where to keep the code. For now it is in /misc, but I'm not sure if that's where it should stay.

misc/B0_multiecho_linfit.m
```diff
@@ -0,0 +1,81 @@
+function [delf] = B0_multiecho_linfit(compl_vol, ph_json)
```
jcohenadad: i know this is counter-intuitive from a physics convention standpoint, but from our established convention, functions should be named with snake-case, i.e.: `b0_multiecho_linfit`.

misc/bkgrnd_noise.m
```diff
@@ -0,0 +1,34 @@
+function sigma = bkgrnd_noise(data_vol)
+
```
jcohenadad, line 22: add header to function please

misc/rescalePhaseImage.m
```diff
@@ -0,0 +1,20 @@
+function [ PHASE_OUT ] = rescalePhaseImage( PHASE_IN, invert )
```
jcohenadad: snake-case for function name please: `rescale_phase_image`

misc/threshold_masking.m
```diff
@@ -0,0 +1,10 @@
+function mask = threshold_masking(mag_data, sigma)
+
+mask = zeros(size(mag_data));
+mask = reshape(mask,size(mag_data,1)*size(mag_data,2)*size(mag_data,3)*size(mag_data,4),1);
+data = reshape(mag_data,size(mag_data,1)*size(mag_data,2)*size(mag_data,3)*size(mag_data,4),1);
+mask(find(data>30*sigma(1))) = 1;
```
jcohenadad: i would make this hard-coded `30` an input parameter (which could be defaulted to 30)

misc/threshold_masking.m
```diff
@@ -0,0 +1,10 @@
+function mask = threshold_masking(mag_data, sigma)
```
jcohenadad: add header to function please

misc/B0_multiecho_linfit.m
po09i: The function should probably go under +imutils/+b0 and the function renamed to multiecho_linfit.

misc/B0_multiecho_linfit.m
```diff
@@ -0,0 +1,81 @@
...
+% Code adapted from https://github.com/evaalonsoortiz/LLSf_B0mapping
+
+% create magnitude and phase data volumes
+mag_data = abs(compl_vol);
+ph_data = angle(compl_vol);
+
+% get number of echoes and echo times
+numTE = size(mag_data,4); 
+delt = [ph_json.EchoTime];
+
```
po09i: Make sure if numTE is the same as size(delt)

misc/rescale_phase_image.m
```diff
@@ -0,0 +1,20 @@
...
+%
+% PHASE_IN: phase image with value ranging from [-4096:4095]
+% invert: Invert sign (i.e. to convert from left-handed system to
+%         right-handed system (default = 1)
+
+if (~exist('invert', 'var'))
+    invert = 1;
+end
+
+PHASE_OUT = (PHASE_IN / 4096) * pi;
```
gaspardcereza: Wouldn't it be simpler to just do `PHASE_OUT = (PHASE_IN / 4096) * pi*invert` thus avoiding the if loop ?

+imutils/+b0/multiecho_linfit.m
```diff
@@ -0,0 +1,81 @@
+function [delf] = multiecho_linfit(compl_vol, ph_json)
+% B0_multiecho_linfit Computes B0 fieldmaps based on a least-squares
```
gaspardcereza: The function name is different from the name in the description

+imutils/+b0/multiecho_linfit.m
```diff
@@ -0,0 +1,81 @@
...
+%      phase json sidecar struct
+%
+% _OUTPUTS_
+%
+%   delf 
+%     field map in units of [Hz]
+% 
+% Code adapted from https://github.com/evaalonsoortiz/LLSf_B0mapping
+
+% create magnitude and phase data volumes
```
gaspardcereza, line 25: I think we decided to go with camelCase for the variable names (magData and phData). @po09i @rtopfer can you confirm ?

misc/rescale_phase_image.m
evaalonsoortiz: In this case invert is a boolean, so invert = 0 means "do not invert the sign". If we wanted to go with your suggestion, then invert would be +1 if you do _not_ want to invert, and -1 if you do. Either way, I've decided to simplify this function so that it works for Siemens data only and never inverts the sign. It is just a temporary function that I was using because I am having issues with [convert_siemens_phase](https://github.com/shimming-toolbox/shimming-toolbox/blob/782fcd6251bdb19f435c27ec10b7a453de98a086/%2Bimutils/read_nii.m#L91).

misc/threshold_masking.m
```diff
@@ -0,0 +1,40 @@
+function mask = threshold_masking(magData, sigma, varargin)
+% Create a mask for an imput image by thresholdoing the signal
```
jcohenadad:
```suggestion
% Create a mask for an input image by thresholding the signal 
```

misc/threshold_masking.m
```diff
@@ -0,0 +1,40 @@
...
+%    sigma
+%      standard deviation of the noise
+%
+%    optional input: signalThreshold
+%      if no value is given signalThreshold = DEFAULT_signalThreshold   
+%
+% _OUTPUTS_
+%
+%   mask 
+%     5D image mask(x,y,z,1,1)
```
jcohenadad: why simply not outputing a mask with the same dimensions as the input image? i find it confusing to return a 5d image (ie not expected behaviour)

evaalonsoortiz: I've clarified in the description that, while the input image can be 5D (x,y,z,t,acq), only the first echo of the first acquisition will be used to generate a mask. So the mask will be 5D with (x,y,z,1,1).

https://github.com/shimming-toolbox/shimming-toolbox/blob/4624469c5f08e1fe18d95e7adaa082881a0af1dc/misc/threshold_masking.m#L1-L4


https://github.com/shimming-toolbox/shimming-toolbox/blob/4624469c5f08e1fe18d95e7adaa082881a0af1dc/misc/threshold_masking.m#L23-L26

jcohenadad: so, does it mean that if the input image is 2d, 3d or 4d, the output mask is also 4d? (currently it states that output is systematically 5d, which i would find strange if a 2d, 3d or 4d image is input).

jcohenadad: also, what is meant by "t" in "(x,y,z,t,acq)"? based on a [previous discussion](https://github.com/shimming-toolbox/shimming-toolbox/pull/131#issue-435964034) isn't the 4th dimension supposed to refer to nEcho?

misc/threshold_masking.m
```diff
@@ -1,7 +1,7 @@
 function mask = threshold_masking(magData, sigma, varargin)
 % Create a mask for an input image by thresholding the signal 
-% Note: the input image can be a 5D image (x,y,z,t,acq), however only the
-% first echo time of the first acquistion is used for thresholding
+% Note: the input image can be a 4D image (x,y,z,nEcho), however only the
```
jcohenadad: based on line 38, input needs to be 4d ( i would replace can by must, to avoid confusion)

evaalonsoortiz: If a 2D multiecho image is supplied (x,y,nEcho), the output will be mask(x,y,1,nEcho) to be consistent with our code elsewhere. If a single-echo 2D image is supplied, the output will be 2D because matlab drops any trailing dimensions that are 1 (ex: (x,y,1) = (x,y)) (in other words: the code will reshape the mask to have 4D, but any last dimension that is 1 will be dropped by matlab). This also goes for a single echo 3D image (x,y,z,1).

jcohenadad: ok, in that case i would explicit the authorized input, something like this (i let you correct what needs to be corrected):
~~~
INPUTS

  magData
    Image from which the mask will be created. Input dimension can be: (x, y, nEcho) or (x, y, z, nEcho).

OUTPUTS:

  mask
    Mask with dimension: (x, y, z, nEcho). Note: if input image was (x, y, nEcho), the function will add a singleton for the "z" dimension to the output: (x, y, 1, nEcho)
~~~

evaalonsoortiz: Done. I think it's probably clearer now.

jcohenadad (approved):

po09i (approved):

PR #129, opened 2020-06-17
Set up gh-pages with azure pipeline

po09i:
Build documentation on azure on push to master and upload to gh-pages branch.

### DIFF
--- a/generate_doc.m
+++ b/generate_doc.m
@@ -91,7 +91,7 @@ outputPath = './s/docs/contributing/api_documentation';
 % rmdir temp s
 % 
 % rmdir shimming-toolbox s
-rmdir helpDocMd s
+
 disp('done')
 exit;
                        
--- a/generate_doc.m
+++ b/generate_doc.m
@@ -23,6 +23,8 @@
 % Clone doc generation software
 %!git clone https://github.com/shimming-toolbox/helpDocMd.git
 % Add various paths
+pwd
+exit
 addpath(genpath('./helpDocMd/src'))
 addpath(genpath('./s'))
 % cd temp
@@ -91,6 +93,7 @@ outputPath = './s/docs/contributing/api_documentation';
 % rmdir shimming-toolbox s
 rmdir helpDocMd s
 disp('done')
+exit;
                        
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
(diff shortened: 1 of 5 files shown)


po09i:
I was able to deploy gh-pages with nothing in it yet. The run was maunally stopped because matlab did not run, possibly updating.

generate_doc.m
```diff
@@ -0,0 +1,172 @@
...
+cd ..
+
+% Cant rename 's' folder to 'shimming-toolbox' since shimming-toolbox
+% already exists in ..
+% !mv -r s shimming-toolbox
+
+% create new temp folder
+% !mkdir temp
+% Clone doc generation software
+!git clone https://github.com/shimming-toolbox/helpDocMd.git
```
po09i: @kousu wrote on shimming toolbox.org

> All these git clones could be done as git submodules. Faster to run the docs, more reproducible too, though more tedious since you always need to be updating the submodules when their versions change.

> For dependencies I think that's generally a positive feature: it makes sure things can't break under you without you knowing about it.

> For shimming-toolbox itself, which is only downloaded so you can run the generator over it, it's more debateable. Though it is convenient in this way: if you change the API, you can update the repo and retrigger the build all in one step by changing the submodule pointer. Without an explicit link like that you'll need to manually re-run the docs pipeline or make nonsense commits just to get the website updated.

generate_doc.m
```diff
@@ -0,0 +1,172 @@
...
+% Add various paths
+addpath(genpath('./helpDocMd/src'))
+addpath(genpath('./s'))
+% cd temp
+
+%% API doc
+% overwrite shimming-toolbox/docs/contributing/api_documentation/
+!rm -r s/docs/contributing/api_documentation
+!mkdir s/docs/contributing/api_documentation
```
po09i: Nick wrote on shimming-toolbox.org

> Rather than doing this, let's just .gitignore that folder so it is never committed in the first place.

po09i:
@kousu wrote in shimming-toolbox.org repo 

> I would like to isolate the matlab parts to the bare minimum required. The mkdocs part, the git clones and anything else, that should all be in a shell script that calls matlab once just to run helpDocMd; ideally it would run like this:

> git clone --depth 1 https://github.com/shimming-toolbox/shimming-toolbox # or: a submodule

> matlab -r "documentor('shimming-toolbox', 'docs/contributing/api_documentation');exit"

po09i:
> @kousu wrote in shimming-toolbox.org repo
> 
> > I would like to isolate the matlab parts to the bare minimum required. The mkdocs part, the git clones and anything else, that should all be in a shell script that calls matlab once just to run helpDocMd; ideally it would run like this:
> 
> > git clone --depth 1 https://github.com/shimming-toolbox/shimming-toolbox # or: a submodule
> 
> > matlab -r "documentor('shimming-toolbox', 'docs/contributing/api_documentation');exit"

I like the Idea, I think a script will still be necessary though to include appropriate paths, exclude unwanted files, etc.

jcohenadad (approved):

Issue #99, opened 2020-06-04
Packaging the code base as a veritable Matlab toolbox


rtopfer:
For ease of installation, upgrading, better options for integrated Matlab documentation, etc.:
[matlab-toolboxes](https://www.mathworks.com/help/matlab/matlab_prog/create-and-share-custom-matlab-toolboxes.html)

rtopfer:
@kousu @jcohenadad @po09i might be prudent to keep this in mind for the [website](https://github.com/shimming-toolbox/shimming-toolbox/pull/129) and API reconfiguration^.

eg. maybe the site itself should just contain documentation for a *stable* release and *not* rebuild with every merge to master?
MATLAB has functionality to upgrade custom toolboxes but idk if/how that integrates with git/CI.

jcohenadad:
this is where readthedocs is useful, in that it can host the doc for multiple releases and branches. Not sure how we can do that with gh-pages (which only points at a single version).

PR #131, opened 2020-06-17
Simple B0 mapping  and unwrapping functions adapted from the experiment.m script

### LINKED ISSUE #128: Implement a simple B0 mapping function
This function is required for testing purpose related to #102. 
It should allow the user to easily compute B0 fielmaps from magnitude and phase 5D arrays (x, y, z, nEchos, nAcquisitions) and from the header containing the echo times (read from the json file).

gaspardcereza:
That issue aims to adapt the B0 mapping code used in `example/experiment.m` to make it easier to call from anywhere (as part of the  `+imutils` package). 

2 functions are added to the project:
`imutils.unwrap.m` allows to choose the algorithm that will be used for the unwrapping and returns the unwrapped phases as a 5D array (x, y, z, nEcho, nAcq). Currently only handles sunwrap.

`imutils.b0.computeFieldMaps.m` allows to choose the algorithm that will be used for the B0 mapping and returns a 4D array (x, y, z, nAcq) corresponding to the different B0 maps at each acquisition time. Currently only handles the phase difference method with 1 or 2 echoes.

Closes #128


+imutils/+B0/phaseDifferenceMapping.m
```diff
@@ -0,0 +1,62 @@
+function B0FieldMap = phaseDifferenceMapping(mag, phase, phaseJson)
+%phaseDifferenceMapping Compute B0 fieldmaps using 1 or 2 echos
+%     
+%     B0FieldMap = phaseDifferenceMapping(mag, phase, phaseJson)
+% 
+% The inputs `mag`, `phase` and `phaseJson` are outputs of the function
```
jcohenadad: i would explicitly define what is expected from mag, phase and phaseJson. is mag a matrix? if so, what dimension? what type? etc.

jcohenadad: it is also important to respect the convention for docstrings. @rtopfer @po09i is the convention written somewhere (i.e. how to list arguments, etc.)

+imutils/+B0/phaseDifferenceMapping.m
```diff
@@ -0,0 +1,62 @@
+function B0FieldMap = phaseDifferenceMapping(mag, phase, phaseJson)
+%phaseDifferenceMapping Compute B0 fieldmaps using 1 or 2 echos
+%     
+%     B0FieldMap = phaseDifferenceMapping(mag, phase, phaseJson)
+% 
+% The inputs `mag`, `phase` and `phaseJson` are outputs of the function 
+% `load_niftis` (from the `imutils` package). 
+
+% Unwrapping of the phase images (using sunwrap)
```
jcohenadad: unwrapping should not be done here-- it should be part of another function. 
the only thing this function does is phaseDIff B0 mapping (as the name of the function says)

+imutils/+B0/phaseDifferenceMapping.m
```diff
@@ -0,0 +1,62 @@
...
+disp('Unwrapping...')
+% Init Unwrapped Phase
+for iAcq = 1:size(phase,5)
+    for iEcho = 1:size(phase,4)
+        % Get the magnitude for a perticular echo
```
jcohenadad:
```suggestion
        % Get the magnitude for a specific echo
```

+imutils/+B0/phaseDifferenceMapping.m
```diff
@@ -0,0 +1,62 @@
...
+        % Calculate the phase in radians, assumes there are wraps
+        phasePi = mat2gray(phase(:,:,:,iEcho,iAcq))*2*pi - pi;
+        
+        % Unwrap phase using sunwrap
+        unwrappedPhase(:,:,:,iEcho,iAcq) = sunwrap(magNorm .* exp( 1i* phasePi ), 0.1);
+        
+    end
+end
+
+% Plot
```
jcohenadad: plot should be the responsibility of another function (to prevent code duplication if we have other B0 mapping functinos)

+imutils/+B0/phaseDifferenceMapping.m
```diff
@@ -0,0 +1,62 @@
...
+B0FieldMapPlot = reshape(B0FieldMap, [size(B0FieldMap, 1) size(B0FieldMap, 2) 1 size(B0FieldMap, 3)]); % montage insists on the format M-by-N-by-1-by-K
+figure(2)
+montage(B0FieldMap,'DisplayRange',[min(B0FieldMapPlot,[],'all') max(B0FieldMapPlot,[],'all')])
+hold on
+colorbar
+title('B0FieldMap (Hz)')
+hold off
+
```
jcohenadad: same comment about the plot

+imutils/+B0/phaseDifferenceMapping.m
```diff
@@ -0,0 +1,62 @@
...
+% Different process if only 1 echo
+if size(unwrappedPhase,4) == 1
+    echoTimeDiff = phaseJson(1).EchoTime;
+    phaseDiff    = unwrappedPhase(:,:,:,:);
+else
+    echoTimeDiff = phaseJson(2).EchoTime - phaseJson(1).EchoTime;
+    % if using wrapped phase % phaseDiff = angle( wrappedPhase(1) .* conj(wrappedPhase(2) ) ) ; then unwrap
+    phaseDiff    = unwrappedPhase(:,:,:,2,:) - unwrappedPhase(:,:,:,1,:);
+end
+
```
jcohenadad: we can probably avoid code duplication and use the same code for 1 and 2 echoes

+imutils/+B0/phaseDifferenceMapping.m
```diff
@@ -0,0 +1,62 @@
...
+% Plot
+B0FieldMapPlot = reshape(B0FieldMap, [size(B0FieldMap, 1) size(B0FieldMap, 2) 1 size(B0FieldMap, 3)]); % montage insists on the format M-by-N-by-1-by-K
+figure(2)
+montage(B0FieldMap,'DisplayRange',[min(B0FieldMapPlot,[],'all') max(B0FieldMapPlot,[],'all')])
+hold on
+colorbar
+title('B0FieldMap (Hz)')
+hold off
+
+disp(['-----'])
```
jcohenadad: those printouts should be avoided in API functions (only use for CLIs)

+imutils/+B0/computeFieldMaps.m
```diff
@@ -0,0 +1,33 @@
...
+    case 'phaseDifference'
+        disp('Computing phase difference B0 maps...')
+        
+        % Different process if only 1 echo
+        if size(unwrappedPhase,4) == 1
+            echoTimeDiff = phaseJson(1).EchoTime;
+            phaseDiff    = unwrappedPhase(:,:,:,:);
+        else
+            echoTimeDiff = phaseJson(2).EchoTime - phaseJson(1).EchoTime;
+            % if using wrapped phase % phaseDiff = angle( wrappedPhase(1) .* conj(wrappedPhase(2) ) ) ; then unwrap
```
po09i: This comment is not needed, it's basically complex division the implementation of complex division.

+imutils/+B0/computeFieldMaps.m
```diff
@@ -0,0 +1,33 @@
+function B0FieldMaps = computeFieldMaps(mappingAlgorithm, unwrappedPhase, phaseJson)
+%computeFieldMaps Computes B0 fieldmaps following the specified algorithm
```
po09i: I'm being finicky but I would assume the input order of the parameters to be : computeFieldMaps(unwrappedPhase, phaseJson, mappingAlgorithm)

+imutils/unwrap.m
```diff
@@ -0,0 +1,34 @@
+function unwrappedPhase = unwrap(unwrapAlgorithm, mag, phase)
+%unwrap Computes the unwrapped phase following the specified algorithm
```
po09i: unwrap(phase, mag, unwrapAlgorithm)?

+imutils/unwrap.m
```diff
@@ -0,0 +1,34 @@
+function unwrappedPhase = unwrap(unwrapAlgorithm, mag, phase)
```
jcohenadad: why not putting it under +B0/?

+imutils/+B0/computeFieldMaps.m
```diff
@@ -0,0 +1,33 @@
...
+%
+%     B0FieldMaps = computeFieldMaps(mappingAlgorithm, unwrappedPhase, phaseJson)
+%
+% The input `mappingAlgorithm` specifies the way the B0 maps are computed.
```
jcohenadad: i believe this is not the proper way of documenting the docstrings, for automatic parsing-- @rtopfer where can we find a template/guideline for docstrings convention?

+imutils/+B0/computeFieldMaps.m
```diff
@@ -0,0 +1,33 @@
...
+        
+        % Different process if only 1 echo
+        if size(unwrappedPhase,4) == 1
+            echoTimeDiff = phaseJson(1).EchoTime;
+            phaseDiff    = unwrappedPhase(:,:,:,:);
+        else
+            echoTimeDiff = phaseJson(2).EchoTime - phaseJson(1).EchoTime;
+            % if using wrapped phase % phaseDiff = angle( wrappedPhase(1) .* conj(wrappedPhase(2) ) ) ; then unwrap
+            phaseDiff    = unwrappedPhase(:,:,:,2,:) - unwrappedPhase(:,:,:,1,:);
+        end
```
jcohenadad: this should go under another function, with its own unit testing.

+imutils/unwrap.m
```diff
@@ -0,0 +1,34 @@
...
+                magNorm = mat2gray(mag(:,:,:,iEcho,iAcq));
+                
+                % Calculate the phase in radians, assumes there are wraps
+                phasePi = mat2gray(phase(:,:,:,iEcho,iAcq))*2*pi - pi;
+                
+                % Unwrap phase using sunwrap
+                unwrappedPhase(:,:,:,iEcho,iAcq) = sunwrap(magNorm .* exp( 1i* phasePi ), 0.1);
+                
+            end
+        end
```
jcohenadad: same comment as above, this will go under its own function

+imutils/unwrap.m
```diff
@@ -0,0 +1,34 @@
...
+    otherwise
+        disp(['Unknown algorithm. The available algorithms are:' newline '- sunwrap'])
+end
+disp('Unwrapping done')
```
jcohenadad: add carriage return at EOF

+imutils/unwrap.m
```diff
@@ -0,0 +1,34 @@
...
+                % Calculate the phase in radians, assumes there are wraps
+                phasePi = mat2gray(phase(:,:,:,iEcho,iAcq))*2*pi - pi;
+                
+                % Unwrap phase using sunwrap
+                unwrappedPhase(:,:,:,iEcho,iAcq) = sunwrap(magNorm .* exp( 1i* phasePi ), 0.1);
+                
+            end
+        end
+    otherwise
+        disp(['Unknown algorithm. The available algorithms are:' newline '- sunwrap'])
```
jcohenadad: available algorithms should be listed in a dict (for example) and automatically fetched from here (to avoid hard-coding it-- prone to error)

+imutils/unwrap.m
jcohenadad: this is going out of the scope of this PR-- you should rename your PR and update the body of your PR to be clear about what this PR covers (in the future i suggest you open another PR). Good habits: the smaller the PR, the better it is, and PRs should be focused on one thing.

+imutils/unwrap.m
jcohenadad: also add unit tests of course

+imutils/+B0/computeFieldMaps.m
rtopfer: Agreed: *mandatory* arguments need to come first (`mappingAlgorithm` can be assigned a default, so it would come later)

More generally, the current form of the arguments doesn't really make sense to me since `phaseJson` is a struct, but you're only using a single field from it—the echo time. We should choose one or the other:
- Either limit the inputs to precisely those that are relevant (when you call the function, `phaseJson.EchoTime` is the argument rather than the entire struct); 
- Or, we decide on a custom data type (create a new `classdef`) for images that bundles the metadata (`phaseJson`) with its image data (`unwrappedPhase`) and stick to it across the code base (i.e. this could be `MaRdI`-*like* but much,much simpler for the sake of maintainability)
(I see pros and cons to both, but obviously for now it's much easier to go with the first option!)

Other points, re: style:

- `camelCase` is for variables: `B0FieldMaps` --> `b0FieldMaps` 
(though `field` is adequate/better IMO since it's shorter and the context is clear)
- the existing code base uses `allowercase()` for functions but we're going to use `snake_case()` from now on: `computeFieldMaps` --> `compute_field_map`

+imutils/+B0/computeFieldMaps.m
rtopfer: Looks ok to me? There's no parsing involved in the documentation: it basically just removes the % sign in the .m header comment and saves it as .md.

Still working out what the guideline really should be for the documentation since as soon as you start using things like the single/triple-` ``` ` when you query `help` or `doc` *in MATLAB* it doesn't reformat any of it, so the display is really ugly. (Hence, the indentation is definitely preferable when possible)

+imutils/+B0/phaseDifferenceMapping.m
rtopfer: Unfortunately Matlab really doesn't have a convention so we have to choose one:
I've just been going by trial and error:
e.g. To me, this looks reasonable *both* in [MATLAB](https://github.com/shimming-toolbox/shimming-toolbox/blob/master/%2Bimutils/read_nii.m) (when you call `help imutils.read_nii) *and* when once it's finally rendered [on the site](https://www.shimming-toolbox.org/contributing/api_documentation/%2Bimutils/read_nii/)

jcohenadad: this seems out-of-scope, so i've opened an issue here, for further discussions: https://github.com/shimming-toolbox/shimming-toolbox/issues/133

+imutils/unwrap.m
rtopfer: since most unwrapping algorithms will require both mag+phase in some form, with additional optional parameters i'd suggest *insisting on a complex input* and get the phase from the single complex argument, e.g. `angle(img)`

+imutils/unwrap.m
rtopfer: does unwrapping ever need to be done for anything other than b0 field mapping? 🤔

jcohenadad: I think so. If at some point I am proven wrong, we can easily put it back where it belongs better, but for now unwrapping for B0 makes more sense to me

mathieuboudreau:
> does unwrapping ever need to be done for anything other than b0 field mapping? 🤔

Yes, there are phase-based B1 mapping techniques (e.g., Bloch-Siegert shift).

+imutils/unwrap.m
jcohenadad: Based on Mathieu’s comment we can leave it where it is now, however I would give it a more explicit name, eg: `unwrap_phase()`

+imutils/+B0/computeFieldMaps.m
jcohenadad: > camelCase is for variables: B0FieldMaps --> b0FieldMaps
(though field is adequate/better IMO since it's shorter and the context is clear)
the existing code base uses allowercase() for functions but we're going to use snake_case() from now on: computeFieldMaps --> compute_field_map

is that documented somewhere in our guidelines?

+imutils/unwrap.m
gaspardcereza: @rtopfer we directly get the mag and phase arrays by loading 2 different niftis folders. So if we want the input to be a complex array we would have to add a step that builds that complex. Though I admit that having a complex array as input is more elegant and makes the function easier to call, I'm thus not sure what solution is the best to make the code clearer.

rtopfer: IMO, it simplifies the code (1 input instead of 2) while tacitly imposing a rule (`mag` always accompanies `phase`). The separate load doesn't matter: I'd just suggest calling the function with an inline operation, e.g.
```
unwrap( mag .* exp( 1i*phase), unwrapAlgorithm );
```

gaspardcereza: Alright it indeed seems simpler. Will do, thanks !

+imutils/unwrap.m
rtopfer: probably better if `unwrapAlgorithm` is provided as a function-handle, e.g.:
```
unwrap( mag .* exp( 1i*phase ), @sunwrap, options )
```

gaspardcereza: I didn't know about function handles. I'll dig into that it seems interesting. But is it possible to fetch it to provide the list of the available functions to the user as Julien proposed?

rtopfer: Not 100% sure but I think it could work as an alternative to creating/maintaining a 'dict', e.g. all the unwrapping options we include in the code base could simply be .m-files in a folder (even if the functions themselves, like misc/prelude.m, only call external programs)
in this way, the menu of options is clear (they're together in one place) and when a user calls the field mapping function, they just give the name of the .m file they want to use prefixed by `@`. 
`options` can be generically whatever the particular unwrapping method needs. 

argument validation / a simple assertion can be added right at the beginning to test whether `unwrapAlgorithm` actually exists, e.g. 
```
assert( isfile( func2str( unwrapAlgorithm ) ) )
```
(and it should throw an error listing the available options (e.g. `what('imutils.unwrappers')`) when it fails)

(obviously, checking for the existence of a file isn't the same as checking that it works—which would be slightly more involved for calls to external programs—but testing *that* is out of the scope + falls partly under issue #99 )

rtopfer: Main point: ideally, it would be cool to keep the code flexible/extensible while still imposing some structure. Using function handles could be one way (a user could even call the field mapping routine with a handle to their own unwrapper, external to the toolbox)

+imutils/unwrap.m
jcohenadad: what about unwrapping algorithms that do not take magnitude as input? at the end, the object of interest is the phase, not the magnitude. So i would make the first arg mandatory (phase), and the second optional (mag)

jcohenadad: also, i find that having a nested operation in the call is not very intuitive (here: creating complex data)

+imutils/unwrap.m
jcohenadad: i would be open to use function handle as input argument. Let's try and see if that fits our needs in the long term (in terms of code clarity, maintenance, etc.)

+imutils/unwrap.m
rtopfer: almost all of them (that i'm aware of) require the magnitude: either directly as a data-quality weights, or to define the region of support (threshold mag -> volume of interest)

if for some reason the magnitude doesn't exist (not sure why that would happen) a pseudo-magnitude could still be provided, e.g. `unwrap( ones(size(phase)) .* exp( 1i*phase ), ... )`

jcohenadad: @gaspardcereza @rtopfer just notifying you in case you did not see my comments above

gaspardcereza: I already implemented the complex argument. 
The user has to be careful because the phase must be in rads (that seems legit as we provide a complex) so we have to make sure that the phase conversion from  [0 4095] to [-pi pi] is done before calling unwrap_phase. 
If ever we find an algorithm that doesn't require the magnitude, it wont be a problem to switch back to the way it was before and right now it indeed seems better to call it with only one argument instead of two.

jcohenadad: - there are algorithms without mag input (Olivier in the lab is currenly working on that)
- i still find it more intuitive to explicitly pass the phase and mag. The code does separate those two anyway, so why not passing them separately?
- Again, i don't like nested operations in calls.

rtopfer: agreed, nested calls should generally be discouraged. i agree with @gaspardcereza that insisting on the complex input has the benefit that it forces the caller to be mindful of the units (another advantage to using a custom data-type: it can carry the units with it). you could also allow different input forms, e.g. 
```
if isreal( varargin{1})
   phase = varargin{1}
else
   mag    = abs(varargin{1})
   phase = angle(varargin{1}) 
end
```

jcohenadad: > we have to make sure that the phase conversion from [0 4095] to [-pi pi] is done before calling unwrap_phase.

to be checked with `assert`, and scenario added to unit test

jcohenadad: > you could also allow different input forms

i would discourage that (at least at this point in our development). We are at a point where we should keep the code and its usage as simple as possible.

+imutils/unwrap.m
gaspardcereza: @rtopfer I implemented the function handles but if I want it to work, i have to call for example:
`B0maps = imutils.B0.field(unwrappedPhase, magJson, @imutils.B0.phase_difference);`
It's now really intuitive but if I only call: 
`B0maps = imutils.B0.field(unwrappedPhase, magJson, @phase_difference);` 
it wont work... One solution would be to put the `phase_difference` function in the same file as `field`. Do you have any other idea of how to make the call looking more like the second example ?

jcohenadad: i'm fine with this:
~~~
B0maps = imutils.B0.field(unwrappedPhase, magJson, @imutils.B0.phase_difference);
~~~

jcohenadad: btw: why inputting the `magJson` (instead of the mag data)?

gaspardcereza: > btw: why inputting the `magJson` (instead of the mag data)?

Here we only need the echo times (that can't be found in phaseJson or magJson). This is not the most intuitive however. I think I'm just gonna ask for the echo times array as input so we will have something like this:
`B0maps = imutils.B0.field(unwrappedPhase, phaseJson.EchoTime, @imutils.B0.phase_difference);`
This way someone who only has the echo times outside of a structure will be able to call the function.

jcohenadad: > Here we only need the echo times

in general, it is good practice to avoid passing physical files in low-level APIs. It makes usage and unit testing more complicated (ie requires a json file). I suggest passing only what is required, ie, echo times.

gaspardcereza: @jcohenadad I tried something: I created 2 sub-folders (`+imutils/+unwrappers` and `+imutils/+B0/+mappers`) that contain the different functions that will be available for unwrapping/mapping. When the argument specifying the function we want to use is different from the available functions, the returned error is:

image

It limits the hard-coding errors because the folder is parsed and the different available functions are gathered and displayed to the user. And the input argument is now only the function name (e.g. `phase_difference`). The only drawback i can think of is that it requires the 2 additional folders mentioned above.

jcohenadad: pls point to the commit so we can all review how the code looks like, thanks

gaspardcereza: The commit: 0d50fbe77374317215fc433f60e3fb4e0fb16c9e

+imutils/+unwrappers/sunwrap.m
```diff
@@ -0,0 +1,36 @@
...
+% _DESCRIPTION_
+%
+% Returns the unwrapped phase maps that have been calculated from the
+% magnitude and phase data at the different echo times and for each
+% acquisition, using the sunwrap Matlab function.
+%
+% _INPUT ARGUMENTS_
+%
+%   complexArray
+%     5D (x,y,z,nEchoes,nAcq) complex array containing the B0 magnitude and
```
jcohenadad: shouldn't we input 4D (x, y, z, nEcho) instead? the temporal aspect is not relevant for this algorithm (ie it will treat each time point independently)

jcohenadad: actually, same comment for the echo: it loops across echoes. The looping should be done in the calling function

rtopfer: it's not clear to me what the purpose of having a `sunwrap` wrapper function is when the source code already exists in '/external'? if it's just for the sake of keeping unwrappers together, why not simply move the original *sunwrap.m*? 

also, if the original *sunwrap.m* only accepts a single echo or time point, bear in mind that you can often avoid `for ` loops using [ `arrayfun` ](https://www.mathworks.com/help/matlab/ref/arrayfun.html). (it depends somewhat on the context which is better, but *usually/when possible* i'd opt for the single line of code with `arrayfun`, rather than at least >3 with `for`)

jcohenadad: > it's not clear to me what the purpose of having a sunwrap wrapper function is when the source code already exists in '/external'? if it's just for the sake of keeping unwrappers together, why not simply move the original sunwrap.m?

+1

gaspardcereza: The idea here was to call all the unwrappers in the same way. I don't know if all the unwrappers will need only one volume at the time (I know there are 3D unwrappers but I don't know if other methods exist). That's why I programmed a new `sunwrap` function. But if all unwrappers are called within a loop (or with arrayFun), then it could be indeed simpler to just move `sunwrap` into the `unwrappers` folder.

jcohenadad: @gaspardcereza to answer this question, could you please list all the unwrappers that we currently have under the project, and if the can all be called with 3d input, let's go with 3d input with the parent function doing the loop.

+unwrap/unwrap_phase.m
```diff
@@ -0,0 +1,55 @@
...
+    availableFunctions = string({meta.package.fromName('unwrap.unwrappers').FunctionList.Name});
+    error(strjoin(['Unwrapping function not found. The available functions are:',...
+        availableFunctions],'\n'));
+end
+
+unwrapper= str2func(['unwrap.unwrappers.' unwrapFunction]);
+
+disp(['Unwrapping the data using ' unwrapFunction])
+
+for iAcq = 1:size(complexArray,5)
```
gab-berest, line 48: I think that if we dig a little bit, there is a way to use a 'along dimension' type of function calling (with possible reshape if necessary). It would be a lot more efficient. However, it depends also on the wrapper/unwrapper implementation. You can look at cellfun maybe.

+b0map/+mappers/phase_difference.m
```diff
@@ -0,0 +1,37 @@
...
+
+if size(unwrappedPhase,4) == 1 % If only 1 echo
+    echoTimeDiff = echoTimes(1);
+    phaseDiff    = unwrappedPhase(:,:,:,:);
+else % 2 echoes mapping
+    echoTimeDiff = echoTimes(2) - echoTimes(1);
+    phaseDiff    = unwrappedPhase(:,:,:,2,:) - unwrappedPhase(:,:,:,1,:);
+end
+
+B0FieldMaps = phaseDiff./(2*pi*echoTimeDiff);
```
gab-berest, line 37: This really doesn't change anything but I think it will be clearer if you do element wise operation on every component of the operation.

+b0map/mapping.m
```diff
@@ -0,0 +1,53 @@
...
+%
+% OUTPUTS
+%
+%   b0FieldMaps
+%     4D (x,y,z,nAcq) array correspondingto the different B0 maps measured
+%     at each acquisition time.
+
+narginchk(2,3)
+
+if nargin == 2 % Check the number of arguments
```
gab-berest, line 35: Since the first two arguments are mendatory, should we add a check to validate these arguments are present and in the correct form since Matlab and users can sometimes do funky things with functions?

gab-berest (commented):
That's what I found ignoring what was already discussed.

+b0map/+mappers/phase_difference.m
```diff
@@ -0,0 +1,37 @@
...
+%     4D (x,y,z,nAcq) array correspondingto the different B0 maps measured
+%     at each acquisition time.
+%
+
+if size(unwrappedPhase,4) == 1 % If only 1 echo
+    echoTimeDiff = echoTimes(1);
+    phaseDiff    = unwrappedPhase(:,:,:,:);
+else % 2 echoes mapping
+    echoTimeDiff = echoTimes(2) - echoTimes(1);
+    phaseDiff    = unwrappedPhase(:,:,:,2,:) - unwrappedPhase(:,:,:,1,:);
```
evaalonsoortiz, line 34: To minimize phase wraps, you should use `atan2(imag(Z1(:,:,:).*conj(Z2(:,:,:))),real(Z1(:,:,:).*conj(Z2(:,:,:))));`
where 
```
Z1(:,:,:) = mag_data(:,:,:,1).*exp(1i*ph_data(:,:,:,1));
Z2(:,:,:) = mag_data(:,:,:,2).*exp(1i*ph_data(:,:,:,2));
```

evaalonsoortiz: My understanding is that once the phase difference is computed, the residual phase wraps can be spatially unwrapped. Here you have the reverse: the phase is first spatially unwrapped, and afterwards you compute the phase difference.

jcohenadad:
@gaspardcereza can we close this?

gaspardcereza:
> @gaspardcereza can we close this?
 
I think since we moved to python some code has been developed based on other B0 mapping technique so I don't think that branch is useful anymore. I guess it can thus be closed.

PR #136, opened 2020-06-18
Update azure-pipeline CI and website generation triggers

### LINKED ISSUE #135: Remove the trigger of Azure pipeline's website generation from PRs
Currently, the website gets built on every push to a PR whose branch was created after June 17th (need to have the azure-pipelines.yml file in the branch). This should be changed so that only the master branch gets built. 

The best solution would be on a merge to master but since azure is built for CI, this might not be available. Other options include scheduled triggers (at noon everyday) and manual triggers before closing a PR (/AzurePipelines run).

po09i:
The purpose of this PR is to set up the trigger the website generation when master is merged and add back the continuous integration on PRs with comments and on merge to master.

Fixes #135

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -18,7 +18,7 @@ steps:
     chmod 755 generate-docs.sh
     ./generate-docs.sh
   displayName: 'Generate and publish website'
-  condition: and(succeeded(),ne(variables['Build.Reason'],'IndividualCI'))
+  condition: and(succeeded(),eq(variables['Build.Reason'],'IndividualCI'))
   env:
     GH_PAGES_TOKEN: $(GH_PAGES_TOKEN)
     CUSTOM_DOMAIN: www.shimming-toolbox.org
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -3,30 +3,19 @@
 trigger:
 - master
 
-# Specifying nothing is the same as this for PRs:
-# pr:
-# - *
+# If there is a push on a PR, it wont do run
+pr: none
 
 # use the self-hosted runner at tristano.neuro.polymtl.ca
 # this has Matlab installed and a license to use it.
 pool: 'Default'
 
 steps:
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -1,26 +1,32 @@
...
+# Always do this step
+- script: |
     # Beware: you MUST put `quit` at the end; Matlab does not auto-quit.
     # And worse, our Matlab license only allows one process at a time (TODO: verify this),
     # so a forgotten `quit` will block all tests on all branches.
     # https://www.mathworks.com/matlabcentral/answers/523194-matlab-script-in-batch-from-unix-command-line
-#    matlab -nodisplay -nosplash -r "disp('Unit tests!'); quit"
-#  displayName: 'Unit Tests'
+    matlab -nodisplay -nosplash -r "disp('Unit tests!'); quit"
+  displayName: 'Unit Tests'
```
kousu: I'd make unit tests a separate pipeline, or at least a separate job in this pipeline, since tests should run on each push to a branch and PR, while updating the website should only happen on pushes to `master`.

kousu: We can make separate pipelines with separate .yml files in Azure:

Screenshot_2020-06-18_13-55-02

po09i: @kousu I was gonna set it up in the same pipeline but without conditions for the CI and adding a condition to the doc generation to only happen on merges/push to master. Is this good practice or do you still recommend seperating in different jobs? Because I believe the triggers are still going to be the same so the conditions are still going to be required.

po09i: @kousu I think your way is more maintainable in the long run so I'l switch to that.

po09i: @kousu I created a new pipeline, could you go approve it please?

kousu (approved):
This looks good!

jcohenadad:
@po09i good habit to close the branch after merging so we don't end up with stale branches (requires maintenance)

PR #150, opened 2020-06-23
Fix website generation bug when creating virtual environment

po09i:
This PR is to fix a bug where the website generation fails due to an error when creating the virtual environment.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -5,7 +5,7 @@ trigger:
 
 # If there is a push on a PR, it wont do run
 # Temporarily to debug in PR
-# pr: none
+pr: none
 
 # use the self-hosted runner at tristano.neuro.polymtl.ca
 # this has Matlab installed and a license to use it.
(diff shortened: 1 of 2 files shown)


po09i:
The website suddenly stopped being generated for no apparent reason. The code fails when executing this [line](https://github.com/shimming-toolbox/shimming-toolbox/blob/af27879add99bda7b5065d4f283b533edf2e95cb/generate-docs.sh#L33) Reading through the [virtualenv documentation](https://virtualenv.pypa.io/en/latest/installation.html#via-pip) revealed potential solutions. The pip installer suggests to be > v9.0.0 and we only have V8.1.1 on Tristano. Is it possible to upgrade it?

jcohenadad:
as per [this related issue](https://github.com/pypa/virtualenv/issues/1876), pip upgrade would be advisable

po09i:
PR is ready for review, virtualenv was downloaded using pip3 on tristano by agah, removing the call to download using pip fixes the problem.

PR #152, opened 2020-07-01
Numerical model with extended deltaB0 capabilities

### LINKED ISSUE #151: Expand the capabilities of NumericalModel.m
NumericalModel.m currently allows the user to simulate T2* decay in a 2D Shepp-Logan phantom using matlab's `phantom` function. A linear B0 gradient can be added to simulate T2* decay in an inhomogeneous B0 field. 

The capabilities of NumericalModel.m should be expanded so that T2* decay is simulated 

- for phantoms of other shapes that will eventually be used in-vivo; such as spherical and cylindrical phantoms

- in the presence of an arbitrary deltaB0 distribution (that can be supplied by the user)

evaalonsoortiz:
This PR will address issue #151. 

@jcohenadad I would like to know if the source code for [phantom3d.m](https://www.mathworks.com/matlabcentral/fileexchange/9416-3d-shepp-logan-phantom) can be included in our toolbox. All that is stated in the [license.txt](https://github.com/shimming-toolbox/shimming-toolbox/blob/numerical_model_with_sus/numerical_model/license.txt) file is that, it (the file) should be included in any distribution of the code. It seems to me that we can. Could you confirm?


numerical_model/NumericalModel.m
```diff
@@ -51,9 +51,8 @@
         end
         
         function obj = shepp_logan(obj, numVox)       
-            % Create a 2D Shepp_Logan volume for the 
-            % dims: [x, y] number of voxels.
-            obj.starting_volume = phantom(numVox);
+            % Create a 3D Shepp_Logan volume with dimensions [numVox numVox numVox]
```
mathieuboudreau: Just a tip - I wrote tests for the expected functionality and so changing this class method from 2D to 3D might make some tests fail. An approach that I had envisioned would be to create different class methods for different phantom models, so in this case I would recommend to create a new method called `shepp_logan_3D` customize your phantom in there. You'd need to also customise the class initialisation to accept a new flag to call that function here:

https://github.com/shimming-toolbox/shimming-toolbox/blob/bf3296ea08605f894712c673a7712dec60e9f817/numerical_model/NumericalModel.m#L38-L43

evaalonsoortiz: Thanks for the tip!

jcohenadad:
> @jcohenadad I would like to know if the source code for [phantom3d.m](https://www.mathworks.com/matlabcentral/fileexchange/9416-3d-shepp-logan-phantom) can be included in our toolbox. All that is stated in the [license.txt](https://github.com/shimming-toolbox/shimming-toolbox/blob/numerical_model_with_sus/numerical_model/license.txt) file is that, it (the file) should be included in any distribution of the code. It seems to me that we can. Could you confirm?

this code is distributed under the [LGPL3 license](https://www.gnu.org/licenses/gpl-3.0.html), stating that _"Everyone is permitted to copy and distribute verbatim copies of this license document, but changing it is not allowed."_, so as long as the code is not modified (and the license+copyright is mentioned, which is the case), then it should be fine.

another consideration, is that LGPL3 is a less permissive license than e.g. MIT, BSD, meaning that our license should be as restrictive (or more) than LGPL3. Fortunately, [shimming-toolbox license](https://github.com/shimming-toolbox/shimming-toolbox/blob/master/LICENSE) is also LGPL3, so no problem here.

evaalonsoortiz:
This PR is now ready to be reviewed. 

NumericalModel.m can now be used to generate simulated T2* decay data for a cylindrical volume, a spherical volume, a 2D or 3D Shepp-Logan phantom, or a "modified" Zubal phantom (the orignial zubal phantom can be found here: http://noodle.med.yale.edu/zubal/data.htm). 

A linear deltaB0 field can be simulated or loaded from an external file.

numerical_model/NumericalModel.m
```diff
@@ -598,7 +598,8 @@
                     else
                         obj.deltaB0 = imrotate( fliplr( niftiread(params) ) , -90);                   
                     end
-                    obj.deltaB0 = (obj.gamma / (2*pi)) * obj.fieldStrength * 1e6 * obj.deltaB0;  
+
```
jcohenadad, line 601: is that the only change in this PR? (ie added blank line)

evaalonsoortiz: No, most of the NumericalModel.m file is new, but the changes made to it were already merged with master via #163 by accident. I was working with 2 branches simultaneously, and modified NumericalModel.m on that branch instead of this one.

jcohenadad (approved):
since most changes were already merged in a previous branch it is safe to approve this one

PR #153, opened 2020-07-01
Rt/ref shim opt deconstruct classes

rtopfer:
This PR will be part of a series dedicated to refactoring shim-related code in the current Coils subdir.

To break-up the immense refactor and hopefully facilitate code review, this initial PR simply migrates a couple `ShimOpt` methods into standalone functions within a new package.

To review/discuss:

-  subpackage folder name `+b0shim/+compute` 
   - alternatively, it could be something like `+b0shim/+math`
   - proposed `+compute` so that a function call looks like `basis = b0shim.compute.spherical_harmonics( ... )` rather than something like `basis = b0shim.math.compute_spherical_harmonics( ... )`

- documentation format

Note: Code for computing coil induction fields via Biot-Savart can probably be added to this same subpackage (possibly in the same PR...)

(Also, p.s. some of the code is inherited so it doesn't completely follow the style convention)


+b0shim/+compute/siemens_basis.m
```diff
@@ -0,0 +1,160 @@
...
+%
+%   - 1 micro-T/m for *X,Y,Z* gradients (= 0.042576 Hz/mm)
+%   - 1 micro-T/m^2 for 2nd order terms (= 0.000042576 Hz/mm^2)
+%
+% The returned `basis` is thereby in the form of ideal "shim reference maps",
+% ready for optimization.
+%
+% INPUTS
+%
+%    orders
```
jcohenadad: i would like very much to see the type indicated for each variable. This is super useful for people not familiar with the code. I've suggested it in this PR (https://github.com/shimming-toolbox/shimming-toolbox/pull/145/files#diff-f7c3aae9cf7ffff5e623bbb43d946a50), which we should merge at some point.

+b0shim/+compute/siemens_basis.m
```diff
@@ -0,0 +1,160 @@
...
+
+%% ------
+% scaling:
+% 1st order terms yield 1 micro-T of field shift per m (i.e 0.042576 Hz/mm )
+% 2nd order terms yield 1 micro-T of field shift per m^2 (i.e 0.000042576 Hz/mm^2 )
+
+% distance from iso/origin to adopted reference point [units: mm]
+r = [1 1 1 1 sqrt(2) sqrt(2) 1 sqrt(2)] ;
+
+% invert polarity of certain terms:
```
jcohenadad: any refs here would be welcome

+b0shim/+compute/siemens_basis.m
```diff
@@ -0,0 +1,160 @@
...
+sh(:,:,:,[2,3,5,8]) = -sh(:,:,:,[2,3,5,8] ) ;
+
+orders = [1 1 1 2 2 2 2 2] ;
```
jcohenadad, line 168: maybe mention in the header that currently the function only works for up to 2nd SH order, and open a feature issue to cover 2+ orders?

+b0shim/+compute/siemens_basis.m
```diff
@@ -0,0 +1,160 @@
...
+% OUTPUTS
+%
+%    basis
+%      4-D double-precision array of spherical harmonic basis fields
+%
+% NOTES/TODO
+%
+% 1. For now, `orders` is, in fact, ignored: fixed as [1:2]—which is suitable
+% for the Prisma (presumably other Siemens systems as well)—however, the
+% 3rd-order shims of the Terra should ultimately be accommodated too. (Requires
```
jcohenadad, line 53: ah! i'm just seeing this-- scrap [this comment](https://github.com/shimming-toolbox/shimming-toolbox/pull/153/files#r448619602) then. 
i would still open an issue-- much more convenient for project management (than having someone open each .m file and look for "todo" fields)

+b0shim/+compute/siemens_basis.m
rtopfer: since matlab isn't strongly typed, we need a convention to indicate generic variables (e.g. `numeric` to encompass `double`, `single`, `int`, etc.?) 

it's also a bit unfortunate since, for instance, `orders` in this case should be integers — but doesn't need to be *typed* as such, so suggesting the specific type is a bit misleading...

+b0shim/+compute/siemens_basis.m
rtopfer: Hmm... none available, just worked this out myself^

+b0shim/+compute/siemens_basis.m
jcohenadad: discussion moved here: https://github.com/shimming-toolbox/shimming-toolbox/pull/145#issuecomment-652653888

+b0shim/+compute/siemens_basis.m
jcohenadad: in that case i would add something like: "(found empirically)". just for transparency.

+b0shim/+compute/siemens_basis.m
```diff
@@ -0,0 +1,160 @@
...
+% ready for optimization.
+%
+% INPUTS
+%
+%    orders
+%      Degrees of the desired terms in the series expansion, specified as a
+%      vector of non-negative integers (`[0:1:n]` yields harmonics up to n-th
+%      order)
+%
+%    X, Y, Z
```
jcohenadad: these are separate input params (not a single ndarray), so i would do:
~~~
%    X (numeric)
%      X-coordinate in the patient-coordinate system (i.e. DICOM reference, units of mm)
%
%    Y (numeric)
%      Y-coordinate in the patient-coordinate system (i.e. DICOM reference, units of mm)
%
%    Z (numeric)
%      Z-coordinate in the patient-coordinate system (i.e. DICOM reference, units of mm)
~~~

jcohenadad: and to suggest that both 2d or 3d work, i would update the usage above:
~~~
%    basis = siemens_basis( orders, X, Y )
%    basis = siemens_basis( orders, X, Y, Z )
~~~

rtopfer: > and to suggest that both 2d or 3d work, i would update the usage above:
> 
> ```
> %    basis = siemens_basis( orders, X, Y )
> %    basis = siemens_basis( orders, X, Y, Z )
> ```

"2d" refers to computing the basis on a single slice—still need the Z coordinates in either case

jcohenadad: > > and to suggest that both 2d or 3d work, i would update the usage above:
> > ```
> > %    basis = siemens_basis( orders, X, Y )
> > %    basis = siemens_basis( orders, X, Y, Z )
> > ```
> 
> "2d" refers to computing the basis on a single slice—still need the Z coordinates in either case

ha! another good example that expliciting what type/size is expected is important for people not familiar with the code (i.e., me 😅).

rtopfer:
@jcohenadad on [this note](https://github.com/shimming-toolbox/shimming-toolbox/pull/145#issuecomment-654539575), can we merge?

jcohenadad (approved):

po09i (approved):

PR #162, opened 2020-08-19
BUG: resliceimg (MaRdI interpolation) issue #161

### LINKED ISSUE #161: Problem using "resliceimg"
The [realtime_zshimming.m](https://github.com/shimming-toolbox/shimming-toolbox/blob/master/example/realtime_zshim.m) script uses a time series of B0 maps to compute the static and dynamic shim gradients that are needed for z-shimming.  If z-shimming is going to be done with a gradient echo scan, then realtime_zshimming.m requires a sample magnitude gradient echo scan so that the static and dynamic shim gradient images can be resampled to the resolution of the gradient echo scan. 

"realtime_zshim.m" calls `resliceimg` here:
https://github.com/shimming-toolbox/shimming-toolbox/blob/9081cdac6701d7df74c5db109c205655be10d54e/example/realtime_zshim.m#L310

Instead of doing realtime z-shimming with a gradient echo scan, I am attempting to do it with an EPI scan. This means that realtime_zshimming.m will use an EPI scan to resample the static and dynamic shim gradient maps to the resolution of EPI scan. When trying this, I run into the following error:

```
Index exceeds the number of array elements (68992).

Error in MaRdI/resliceimg (line 1342)
            [~,iNearest(iR)] = min( sqrt( ( X_Ip - X_Ep( iEp(iR) ) ).^2 + ...

Error in realtime_zshim (line 310)
GzField.resliceimg( X,Y,Z, mask ) ; % reslice static b0 image
```

There is sample data to test this out [here](https://github.com/shimming-toolbox/data-testing/tree/master/realtime_zshimming_data). The realtime_zshim.m script can be called using `realtime_zshim('phantom')` in the folder containing the sample data. You will then be prompted to enter the paths of the B0 maps and the EPI (or GRE) images.

rtopfer:
PR fixes #161 — bug in MaRdI.m ([constructor](https://github.com/shimming-toolbox/shimming-toolbox/blob/9081cdac6701d7df74c5db109c205655be10d54e/Img/MaRdI.m#L121)) when the number of image volumes acquired for a given series (a.k.a. number of repetitions, or measurements) as stated in the dicom header differs from the number of image files discovered (i.e. provided by the user). 

- `MaRdI()` now issues a warning when such a discrepancy occurs; nevertheless proceeding with object construction.

Also fixed are several erroneous variable assignments/reassignments in [`MaRdI.resliceimg()` ](https://github.com/shimming-toolbox/shimming-toolbox/blob/9081cdac6701d7df74c5db109c205655be10d54e/Img/MaRdI.m#L1150) re: array dimensions.

### DIFF
--- a/Img/MaRdI.m
+++ b/Img/MaRdI.m
@@ -1235,41 +1235,32 @@ elseif nargin >= 5
     end
 end
 
-isUsingScatteredInterpolant = [] ;
-
-if (ndims(Img.img) > 1) && (ndims(Img.img) <= 5)
-
-    gridSizeIp = Img.getgridsize() ;
-    
-    if gridSizeIp(3) > 1
-        isUsingScatteredInterpolant = true ;
-    else
-        isUsingScatteredInterpolant = false ;
-    end
-else
-    error('Dimensions of input Img.img must be >= 2, and <= 5') ;
-end
+assert( [ndims(Img.img) >= 2] & [ndims(Img.img) <= 5],...
+    'Dimensions of input Img.img must be >= 2, and <= 5') ;
 
 % -----------------
 %% Define variables
+gridSizeIp = Img.getgridsize();
+
+if gridSizeIp(3) > 1
+    isUsingScatteredInterpolant = true ;
+else
+    isUsingScatteredInterpolant = false ;
+end
+
 
 switch ndims(X_Ep)
-    case 2  
-        % interpolating down to a single-slice
+    case 2  % interpolating down to a single-slice
         gridSizeEp = [ size(X_Ep) 1 ] ;
-    case 3
-        % interpolating to a 3d volume
+    case 3 % interpolating to a 3d volume
         gridSizeEp = size(X_Ep) ;
-    otherwise
-        % sanity check 
+    otherwise % sanity check 
         error('Expected 2d or 3d target interpolation grid')
 end
 
 % sanity check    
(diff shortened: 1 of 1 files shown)


evaalonsoortiz:
@rtopfer I just tested this out, and was able to use the testing dataset that has a number of volume repetitions that differs from the original number. To confirm, the original error when using` MaRdI.resliceimg() ` still occurs. This is expected, right?

rtopfer:
@evaalonsoortiz should work now? tbh the whole function seems way more complicated than it ought to be... i feel there must be a simpler built-in matlab function somewhere 🧐

evaalonsoortiz:
Hmmm, still getting an error. I am using the EPI image in the sample dataset "34_ep2d_bold_realtime_zshim_UNSHIMMED_2" instead of a GRE for the resampling. The error I am getting now is slightly different:

```
Error using MaRdI/resliceimg (line 1283)
Dimensions of the arrays of extrapolation coordinate and the assigned extrapolation ROI mask must be identical

Error in realtime_zshim (line 310)
GzField.resliceimg( X,Y,Z, mask ) ; % reslice static b0 image
```

evaalonsoortiz:
> @evaalonsoortiz should work now? tbh the whole function seems way more complicated than it ought to be... i feel there must be a simpler built-in matlab function somewhere 🧐

I think ANTs ResampleImage could replace this once the code is fully nifti-functional.

Issue #169, opened 2020-09-29
dual_echo.m computes the phase difference for a right-handed system


evaalonsoortiz:
We mostly work with Siemens, which is a left-handed system, so this is confusing. A "handedness" parameter should be used to switch between the two.

PR #170, opened 2020-09-29
added a handedness variable to switch between left- and right-handed systems

evaalonsoortiz:
solves issue #169


jcohenadad (approved):
awesome! this should also be implemented in the python version @po09i

po09i (approved):

